PR #14, opened 2025-02-06
Cip 0043 Draft

hythloda:


waynecollier-da (changes requested):
I'd like to see a "Specification" and "Rationale" section in this doc, following CIP-0000. The Specification would include the information in the Abstract, the "In the event that TRM fails..." clause, and the Additional Clauses. 

The "Rationale" section would explain why a weight of 2.5 is justified, compared to other SVs that will receive a weight of 1.0 for similar commitments. For example, is Elliptic saying it will not engage with regulators or join the GSF? And is doing those two things (which appear to be the only difference from Elliptic) worth a 1.5 boost in weight?

waynecollier-da (approved):

PR #20, opened 2025-02-17
CIP 0047 Featured App Activity Markers

moritzkiefer-da:


cip-xxxx/cip-xxxx.md
```diff
@@ -0,0 +1,157 @@
+<pre>
+  CIP: Number Pending
```
moritzkiefer-da, line 2: @stas-sbi @waynecollier-da @isegall-da @tkatrichenko I believe you are the current CIP editors, let me know when you assigned a number and I'll update this and the file names.

cip-xxxx/cip-xxxx.md
```diff
@@ -0,0 +1,157 @@
...
+transfers. This can be useful for composed applications like an
+exchange where both an exchange and the asset registry applications should get rewards
+for successfully settling a trade in a single, atomic DvP transaction.
+
+Featured application providers are expected to create featured
+application activity markers only for transactions that correspond to a
+transfer of an asset, or an equivalent transaction, which was 
+enabled by the application provider. The
+detailed fair usage policy and enforcement thereof is left up to the
+Tokenomics Committee of the Global Synchronizer Foundation (GSF).
```
stas-sbi, line 83: I'm wondering how the enforcement would work. What stops an app operator of changing the app after being approved?

moritzkiefer-da: good question!

Featured app rewards (including the existing ones for CC transfer so this isn't a new problem) deliberately allow you to get more rewards than you pay in terms of fees. That inherently allows you to game them in some form.

The way we handle this is by explicitly grating a `FeaturedAppRight` only to vetted applications and that right can be revoked at any point. So you're right that an app operator can change their app, but we can observe that they suddenly get a lot more rewards and other operators can observe they get less (as the total is capped) and they will then have to justify themselves towards the tokenomics committee and their right will get revoked if they are deemed to abuse it.

stas-sbi: Not directly related to this CIP but I'd say that monitoring and revoking is likely to be difficult. Maybe featured apps should have expiration by default instead.

moritzkiefer-da: Interesting idea, so far we've observed that featured app providers monitor their own rewards very closely and did raise issues when they changed so in that regard I'd argue the monitoring process does seem to work pretty well but I agree it's a tricky problem.

cmatturri: How exactly will the 'app activity marker' be measured- will this include some measurement of onchain users connecting to the app?

We are building the Canton Scan, which will become a data/ reporting hub for SVs, Vs, & other applications and can keep this in mind as we roll out our application soon.

moritzkiefer-da: You will be able to observe the creation of the marker contract, the conversion to the AppRewardCoupon and then the minting on the scan update API that you're probably already familiar with (https://github.com/hyperledger-labs/splice/blob/main/apps/scan/src/main/openapi/scan-internal.yaml#L514)

cip-xxxx/cip-xxxx.md
waynecollier-da: @hythloda is also an editor

cip-xxxx/cip-xxxx.md
```diff
@@ -0,0 +1,157 @@
...
+US Dollar amount determined by `featuredAppActivityMarkerAmount` and converted
+to Canton Coin based on the Canton Coin conversion rate associated
+with that mining round.
+
+These `AppRewardCoupon` contracts can then be minted in the same
+fashion as activity records originating from Canton Coin transfers. In
+particular, they are minted from the existing minting pool/tranche for
+application rewards.
+
+One transaction can contain multiple markers for different providers
```
matt-brale-xyz, line 72: Are the markers required to be created in the same transaction as the associated activity?   I think there might be some activity scenarios where the facilitating app won't be the one initiating the transaction that completes the activity.   For example, an app issuing tokens to parties on other participants via the tokenization utility's Offer contracts; if the accepting party is on a different participant from the app, the last transaction will be initiated by the recipient accepting the offer made by the party in the rewarded app via their own participant's API.

moritzkiefer-da: It's not enforced technically but I would expect that this is required by policy.

Your example is an interesting one! However, I think it's important to realize that creating a featured app marker for party X does not require that party X submits the transaction. Daml supports delegation. So in your case I would expect that the offer has the app provider (i.e. the issuer) as a signatory. The Accept choice on that offer can then create the marker contract for that provider even if it is not submitted by the provider but by the recipient of the offer.
This is the sample principle for the current CC transfers: You don't need to be the sender of a CC transfer to get the featured app reward but instead there is a dedicated provider party.

cip-xxxx/cip-xxxx.md
moritzkiefer-da: thank you @hythloda!

waynecollier-da (approved):

cip-0047/cip-0047.md
```diff
@@ -1,7 +1,7 @@
 <pre>
   CIP:  CIP 0047
   Title: Featured App Activity Markers
-  Author: Moritz Kiefer
+  Author: Moritz Kiefer, Simon Meier
```
meiersi-da, line 4: Thx. Note that Wayne stated on my CIP that the authors need to be separated by a newline.

cip-xxxx/cip-xxxx.md
hythloda: this seems resolved.

waynecollier-da (approved):

PR #33, opened 2025-03-04
adapt CIP-0048 to explicitly list futureValues

isegall-da:


waynecollier-da:
This PR corrects an error and adds detail that was missing in the approved version of CIP-0048, to bring the specification in line with the motivation and rationale, and the overall objective of the CIP. 

Instead of applying the new, raised caps to the first phase of the minting curve, it applies them to the current and future phases of the minting curve. 

Please add your comments below.

hythloda (dismissed):

waynecollier-da:
@isegall-da or @hythloda Can you merge this?

waynecollier-da (approved):

PR #35, opened 2025-03-04
first commit of a CIP for changes to on-chain governance votes

isegall-da:


isegall-governance-changes/cip-XXXX.md
```diff
@@ -0,0 +1,97 @@
...
+  Comments-Summary: No comments yet
+  Status: Draft
+  Type: Governance
+  Created: 2025-03-04
+  License: CC0-1.0
+</pre>
+
+## Abstract
+
+As the Super Validators developed experience with on-ledger voting, we realized that the current
```
waynecollier-da, line 16: This should go into the Motivation section. 

For this section I'd just state your a) to e) points, without embellishment.

isegall-governance-changes/cip-XXXX.md
```diff
@@ -0,0 +1,97 @@
...
+a) votes on config changes will change individual (or a few) config values, and not the whole set of configurations;
+b) vote proposals will have an expiration date, after which if not enough votes have been submitted,
+they will be auto-rejected;
+c) votes will remain open for editing until they become effective;
+d) votes will be early-rejected upon 2/3 majority of reject votes;
+e) vote requests can be labeled as "immediate effectivity",
+meaning they will automatically get accepted and take effect immediately as soon as 2/3 of the
+Super Validator nodes vote in favor.
+
+## Copyright
```
waynecollier-da, line 29: Copyright should move to the bottom of the doc; sorry if the template is out of date.

isegall-governance-changes/cip-XXXX.md
```diff
@@ -0,0 +1,97 @@
...
+
+This CIP is licensed under CC0-1.0: [Creative Commons CC0 1.0 Universal](https://creativecommons.org/publicdomain/zero/1.0/)
+
+## Specification
```
waynecollier-da, line 33: Specification should just describe what the operator needs to check against the Daml code to confirm that it does what the Specification describes. 

The rest of this reads like Rationale; meaning, why is this solution the best approach to solve the problem stated in the Motivation.

isegall-governance-changes/cip-XXXX.md
```diff
@@ -0,0 +1,97 @@
...
+expiration date. This would allow taking relatively quick actions, without having to wait after the
+expiration date, and without having to put much thought into chosing an effectivity date when
+creating the vote request.
+
+
+## Motivation
+
+The changes propsed here are all improvements to the operational governance process, based on
+the collective experience that the Super Validator operators have developed over the past months.
+
```
waynecollier-da, line 84: Motivation should be the problem statement: background, description of the problem specifics, and why this needs to be solved now.

isegall-governance-changes/cip-XXXX.md
```diff
@@ -0,0 +1,97 @@
...
+## Backwards compatibility
```
waynecollier-da, line 85: mIssing a section on Rationale, aka solution description. Mostly the stuff you've already explained in the Specification section above.

isegall-governance-changes/cip-XXXX.md
```diff
@@ -0,0 +1,97 @@
...
+the change.
+
+### Expiration as a Due Date for Enough Votes
+
+Every vote request will have an expiration date. This date should be thought of as the due date for
+Super Validator operators to act on this vote request. If at the time of expiration, not enough
+votes are in (i.e. less than 2/3 of the Super Validators have put in any vote), then the vote request
+is immediately considered stale and auto-rejected.
+
+### Votes Remain Open
```
stas-sbi, line 53: - What purpose would expiration date serve when it doesn't affect ability to change votes?
- What happens new votes (parties which haven't voted by the time) after expiration date, can they still vote?

waynecollier-da: The purpose of the expiration date is to close out the vote if no one / too few SVs vote on it by that date. 

We're not trying to gate it, just discard it.

isegall-governance-changes/cip-XXXX.md
```diff
@@ -0,0 +1,97 @@
...
+meaning they will automatically get accepted and take effect immediately as soon as 2/3 of the
+Super Validator nodes vote in favor.
+
+## Copyright
+
+This CIP is licensed under CC0-1.0: [Creative Commons CC0 1.0 Universal](https://creativecommons.org/publicdomain/zero/1.0/)
+
+## Specification
+
+### Vote on Individual Configuration Values
```
stas-sbi, line 35: Can I suggest to move this out to a separate CIP?

isegall-governance-changes/cip-XXXX.md
stas-sbi: Would it be still useful when early rejection is implemented? I also think that expiration date is somewhat non-obvious. Until today I thought that the only purpose for it was to freeze voting. If it's still useful an we want to keep it then we'd need to make it more clear on the UI what it means.

isegall-governance-changes/cip-XXXX.md
```diff
@@ -0,0 +1,97 @@
...
+
+In the current implementation, every vote on config changes (e.g. the AmuletConfig, which governs
+mining, fees, etc) replaces the entire set of configuration values from the current set to those
+specified in the vote request. This has often caused confusion when multiple vote requests were
+created roughly at the same time, but their interaction with each other was hard to reason about,
+ending up with conflicts, and with votes that accidentally overwrote each other's values. In the
+new implementation, a vote request will only apply changes to those fields that the proposer has
+chosen to change, leaving the others unmodified, regardless of their value at the time of applying
+the change.
```
stas-sbi, line 44: Would the changed fields be locked by a vote proposal?

isegall-governance-changes/cip-XXXX.md
moritzkiefer-da: early rejection doesn't handle the case where you just don't get enough votes to reject it early.

isegall-governance-changes/cip-XXXX.md
moritzkiefer-da: The UI will reject two concurrent vote proposals modifying the same fields yes

isegall-governance-changes/cip-XXXX.md
```diff
@@ -0,0 +1,97 @@
...
+after these changes are applied, and leave past votes intact. The UI of the SV application will
+be adapted to match the new logic, but will come into use in sync with the changes in the Daml
+models described here.
+
+## Reference implementation
+
+We have started implementing the required changes for this CIP, which include changes in the
+governance Daml models, as well as the backend and frontend of the SV application . Access to a test
+environment will be provided to all Super Validator Operators to experiment with and comment on
+before we finalize the code and finalize the changes.
```
moritzkiefer-da, line 97: can we link to a branch containing the Daml changes?

isegall-governance-changes/cip-XXXX.md
isegall-da: I don't have an opinion on that, I'd go with whatever approach increases the chances of a quick process toward approval. Do you think a separate CIP will improve that?

isegall-da: @waynecollier-da WDYT?

isegall-governance-changes/cip-XXXX.md
isegall-da: > Until today I thought that the only purpose for it was to freeze voting

That was indeed the semantics so far. We propose to change that, and make it only a deadline for putting some vote in.

isegall-governance-changes/cip-XXXX.md
isegall-da: Note that the Daml changes are intertwined, so from the perspective of reviewing the Daml code, I think a single CIP makes it easier.

isegall-governance-changes/cip-XXXX.md
isegall-da: We can, but I thought the CIP is about the intent, not about the implementation. Linking to the code might later raise objections if we still need to modify that code after the CIP is approved (in a way that is still compliant with the textual description in the CIP)

isegall-governance-changes/cip-XXXX.md
waynecollier-da: @stas-sbi Yes: we're reducing the semantics of "expiration". It is now literally just a date when the vote proposal will expire if it is ignored, like food expiring on a shelf. 

We decided it was a bad idea to freeze voting, given that we've seen cases where someone needed to correct their vote late. Freezing voting also made it harder for organizations to get credit for their vote when they had a longer approval process. 

Early rejection is the opposite of expiration: it's a conscious choice to quickly throw the proposal out.

isegall-governance-changes/cip-XXXX.md
waynecollier-da: I don't want to put this in a separate CIP, because:

1. The Daml models are highly related to each other, and
2. It will be hard to get two related CIPs through the voting process; the SV node owners will want to see just one.

isegall-governance-changes/cip-XXXX.md
```diff
@@ -0,0 +1,97 @@
...
+
+## Backwards compatibility
+
+Vote requests are short-lived. The changes proposed here will apply only to vote requests created
+after these changes are applied, and leave past votes intact. The UI of the SV application will
+be adapted to match the new logic, but will come into use in sync with the changes in the Daml
+models described here.
+
+## Reference implementation
+
```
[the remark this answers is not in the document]
waynecollier-da: ok, not the test environment then.

isegall-governance-changes/cip-XXXX.md
waynecollier-da: @isegall-da The process assumes the code will be developed separately, and the node operators will make the final call whether it is in line with the CIP.  We should link the branch.

isegall-governance-changes/cip-XXXX.md
```diff
@@ -0,0 +1,122 @@
...
+be rejected.
+
+### Immediate Effectivity
+
+Currently, any vote request has both an expiration date and effectivity date.
+In practice, we've seen that on many vote proposals, the intention is to apply then
+as soon as possible and not at a pre-determined future date. Supporting "immediate effectivity"
+will satisfy this requirement.
+
+## Motivation
```
waynecollier-da, line 98: Motivation goes before Rationale

isegall-governance-changes/cip-XXXX.md
```diff
@@ -0,0 +1,122 @@
...
+  Status: Draft
+  Type: Governance
+  Created: 2025-03-04
+  License: CC0-1.0
+</pre>
+
+## Abstract
+
+This CIP proposes the following changes to on-chain voting:
+a) votes on config changes will change individual (or a few) config values, and not the whole set of configurations;
```
waynecollier-da, line 17: Add line breaks between list elements

isegall-governance-changes/cip-XXXX.md
```diff
@@ -0,0 +1,127 @@
...
+
+The new implementation will introduce an "immediate effectivity" option for vote requests,
+which does not specify an explicit effectivity date, but instead becomes effective immediately
+upon collecting 2/3 "Accept" votes, but not earlier than the expiration date.
+
+## Motivation
+
+As the Super Validators developed experience with on-ledger voting, we realized that the current
+implementation is too complex, error-prone, and hard to reason about. We've encountered multiple
+cases where votes with mistakes have been created, that could have been avoided with more
```
waynecollier-da, line 65: edit for clarity: 

...where vote proposals with mistakes have been submitted. These mistakes could have been avoided...

isegall-governance-changes/cip-XXXX.md
```diff
@@ -0,0 +1,127 @@
...
+### Expiration as a Due Date for Enough Votes
+
+Every vote request will have an expiration date. If at the time of expiration, not enough
+votes are in (i.e. less than 2/3 of the Super Validators have put in any vote), then the
+vote request is immediately considered stale and auto-rejected.
+
+### Votes Remain Open
+
+The new implementation will allow votes to be modified until the point where the action being
+vote on has become effective.
```
waynecollier-da, line 48: action being voted on is scheduled to go into effect.

isegall-governance-changes/cip-XXXX.md
```diff
@@ -0,0 +1,127 @@
...
+### Early Rejection
+
+In the new implementation, vote requests that collected 2/3 "Reject" votes will immediately
+close as rejected.
+
+### Immediate Effectivity
+
+The new implementation will introduce an "immediate effectivity" option for vote requests,
+which does not specify an explicit effectivity date, but instead becomes effective immediately
+upon collecting 2/3 "Accept" votes, but not earlier than the expiration date.
```
waynecollier-da, line 59: Wait, I don't understand this last point "but not earlier than the expiration date". 

Immediate effectivity doesn't care about the expiration date.  It's not good behavior to use immediate effectivity to shut out organizations that have slow voting processes, but on things like weight changes or bug fixes, we should agree to use immediate effectivity. 

If we're going to wait for the expiration date, we should just put the effectivity right after the expiration.

isegall-governance-changes/cip-XXXX.md
waynecollier-da, line 48: voted on

isegall-governance-changes/cip-XXXX.md
```diff
@@ -0,0 +1,127 @@
...
+in the pipeline for longer than they should. Making the expiration date the deadline for
+operators to act upon the vote will help operators keep track of their required actions and
+when they are required, as well as clean up stale votes that failed to get attention.
+
+### Votes Remain Open
+
+In the current implementation, once the expiration date on a vote request has passed,
+votes on it can no longer be modified. In practice, this has proven to be a burdening
+limitation, and Super Validator operators often expressed a need to modify their vote
+until the very last minute, i.e. until the corresponding action as become effective.
```
waynecollier-da, line 93: has become effective

isegall-governance-changes/cip-XXXX.md
```diff
@@ -0,0 +1,127 @@
...
+The new implementation will satisfy that requirement by allowing any vote to be modified
+until the point where the vote has become effective.
+
+### Early Rejection
+
+Mistakes happen. We have had several cases on ledger already where vote requests contained
+errors, and we quickly realized they should get rejected. In the current implementation,
```
waynecollier-da, line 100: and the Super Validator operators quickly realized the votes should be rejected.

isegall-governance-changes/cip-XXXX.md
```diff
@@ -0,0 +1,127 @@
...
+In the current implementation, once the expiration date on a vote request has passed,
+votes on it can no longer be modified. In practice, this has proven to be a burdening
+limitation, and Super Validator operators often expressed a need to modify their vote
+until the very last minute, i.e. until the corresponding action as become effective.
+The new implementation will satisfy that requirement by allowing any vote to be modified
+until the point where the vote has become effective.
+
+### Early Rejection
+
+Mistakes happen. We have had several cases on ledger already where vote requests contained
```
waynecollier-da, line 99: vote proposals

isegall-governance-changes/cip-XXXX.md
```diff
@@ -0,0 +1,127 @@
...
+errors, and we quickly realized they should get rejected. In the current implementation,
+there is no way to quickly get rid of those votes despite wide agreement that they should
```
waynecollier-da, line 101: vote proposals

isegall-governance-changes/cip-XXXX.md
```diff
@@ -0,0 +1,127 @@
...
+be rejected.
+
+### Immediate Effectivity
+
+Currently, any vote request has both an expiration date and effectivity date.
```
waynecollier-da, line 106: If a vote passes its expiration date with 2/3 of the SVs in favor, it's highly likely that it will go into effect on chain. For vote proposals that impact application providers and Validator operators, it's helpful to know for certain that a vote will take effect, and then have some time to prepare for this impact. Providing separate expiration and effectivity dates meets this need.

isegall-governance-changes/cip-XXXX.md
```diff
@@ -0,0 +1,127 @@
...
+In practice, we've seen that on many vote proposals, the intention is to apply then
```
waynecollier-da, line 107: On the other hand, some votes only involve the Super Validators, and have little to no operational impact. And there is also the possibility that the Super Validators could want to introduce a bug fix quickly. In these cases, it would be valuable to apply a vote proposal as soon as possible rather than at a pre-determined future date.

isegall-governance-changes/cip-XXXX.md
waynecollier-da: (just asking you to fix the typo "vote on"

waynecollier-da (approved):

waynecollier-da:
@stas-sbi Are you willing to resolve the pending discussions, or do you feel strongly we need to change the plan?

isegall-governance-changes/cip-XXXX.md
stas-sbi: > It will be hard to get two closely related CIPs through the voting process; the SV node owners will want to see just one.

ability to change individual fields is a technical improvement while the rest affects sv operator processes, I'd "vote" for the first one right away while the rest is worth discussing

it splitting these is too much hassle I won't insist but I would cite cip-0000:
> A single CIP should contain a single key proposal or new idea. The more focused the CIP, the more successful it tends to be. If in doubt, split your CIP into several well-focused ones.

isegall-governance-changes/cip-XXXX.md
stas-sbi: If implemented then everyone (unless it's optional? it's not clear from the CIP) would have to be considering the expiration date when creating a new vote. If so:
- The purpose (effect) of this field should be clear from the UI
- There should be some "guidance" on how to decide on this value, maybe there should be a default which we don't have to touch in most cases

isegall-governance-changes/cip-XXXX.md
waynecollier-da: > ability to change individual fields is a technical improvement while the rest affects sv operator processes, I'd "vote" for the first one right away while the rest is worth discussing

> if splitting these is too much hassle I won't insist but I would cite cip-0000:

I appreciate your flexibility. From the perspective of DA's Splice contributors, these are highly intertwined; they are worked on by the same engineers in the same project. 

I think this is also the same for the voters: they will see all of these governance changes as a single, interrelated topic. 

I think it is right to reference the CIP-0000 guidance. I just think it does not lead to the conclusion that we need to separate CIPs in in this specific case, due to the intertwined nature of the technical work.

isegall-governance-changes/cip-XXXX.md
waynecollier-da: Thanks, Stas. We'll make sure the Expiration date has a default value (currently 7 calendar days, based on the votes taken on this topic in the SV Ops meeting). Effectivity will default to two days after Expiration, but it can be moved to take place right after expiration, or moved forward to happen later.

waynecollier-da:
@hythloda Can you assign a number to this CIP and add it to the Index?

waynecollier-da (dismissed):

waynecollier-da (approved):

cip-0051/cip-0051.md
```diff
@@ -0,0 +1,133 @@
...
+## Abstract
+
+This CIP proposes the following changes to on-chain voting:
+
+a) votes on config changes will change individual (or a few) config values, and not the whole set of configurations;
+
+b) vote proposals will have an expiration date, after which if not enough votes have been submitted,
+they will be auto-rejected;
+
+c) votes will remain open for editing until they become effective;
```
waynecollier-da, line 23: change to "votes on a proposal may be changed until the vote proposal becomes effective"

cip-0051/cip-0051.md
```diff
@@ -0,0 +1,133 @@
...
+proposer has chosen to change, leaving the others unmodified, regardless of their value at the
+time of applying the change.
+
+### Expiration as a Due Date for Enough Votes
+
+Every vote request will have an expiration date. If at the time of expiration, not enough
+votes are in (i.e. less than 2/3 of the Super Validators have put in any vote), then the
+vote request is immediately considered stale and auto-rejected.
+
+### Votes Remain Open
```
waynecollier-da, line 45: Change to "Vote Proposals Remain Open"

cip-0051/cip-0051.md
```diff
@@ -0,0 +1,133 @@
...
+
+The new implementation will allow votes to be modified until the point where the action being
+vote is scheduled to go into effect.
```
waynecollier-da, line 48: "voted on"

cip-0051/cip-0051.md
```diff
@@ -0,0 +1,133 @@
...
+
+### Expiration as a Due Date for Enough Votes
+
+Every vote request will have an expiration date. If at the time of expiration, not enough
+votes are in (i.e. less than 2/3 of the Super Validators have put in any vote), then the
+vote request is immediately considered stale and auto-rejected.
+
+### Votes Remain Open
+
+The new implementation will allow votes to be modified until the point where the action being
```
waynecollier-da, line 47: "The new implementation will allow votes on a proposal to be modified until the point where the action being"

cip-0051/cip-0051.md
```diff
@@ -0,0 +1,133 @@
...
+vote is scheduled to go into effect.
+
```
waynecollier-da, line 49: Add:

The text of a proposal is locked once it is submitted and cannot be modified during voting. The proposal can only be accepted or rejected, or it may expire.

cip-0051/cip-0051.md
```diff
@@ -0,0 +1,136 @@
...
+d) votes will be early-rejected upon 2/3 majority of reject votes;
+
+e) vote requests can be labeled as "immediate effectivity",
+meaning they will automatically get accepted and take effect immediately as soon as 2/3 of the
+Super Validator nodes vote in favor.
+
+## Specification
+
+### Vote on Individual Configuration Values
```
hythloda, line 33:
```suggestion
### Make it be like this instead
```

cip-0051/cip-0051.md
```diff
@@ -0,0 +1,136 @@
...
+### Early Rejection
+
+In the new implementation, vote requests that collected 2/3 "Reject" votes will immediately
+close as rejected.
+
+### Immediate Effectivity
+
+The new implementation will introduce an "immediate effectivity" option for vote requests,
+which does not specify an explicit effectivity date, but instead becomes effective immediately
+upon collecting 2/3 "Accept" votes, but not earlier than the expiration date.
```
waynecollier-da, line 62:
```suggestion
upon collecting 2/3 "Accept" votes, but not earlier than the expiration date.
```

waynecollier-da, line 62:
```suggestion
upon collecting 2/3 "Accept" votes.
```

cip-0051/cip-0051.md
```diff
@@ -0,0 +1,136 @@
...
+### Early Rejection
+
+In the new implementation, vote requests that collected 2/3 "Reject" votes will immediately
+close as rejected.
+
+### Immediate Effectivity
+
+The new implementation will introduce an "immediate effectivity" option for vote requests,
+which does not specify an explicit effectivity date, but instead becomes effective immediately
+upon collecting 2/3 "Accept" votes.
```
waynecollier-da, line 62:
```suggestion
The new implementation will introduce an "immediate effectivity" option for vote proposals,
which does not specify an explicit effectivity date, but instead becomes effective immediately
upon collecting 2/3 "Accept" votes.

Vote proposals that use immediate effectivity would be independent of the expiration date, and may happen prior to the expiration date.  

If there is a desire for the vote proposal to either become effective or expire at the same date-time as its expiration, the submitter may set the "effective date" to the same date-time as the "expiration date"
```

waynecollier-da (approved):

PR #73, opened 2025-07-02
CIP-00XX: One-Time Allocation of Historical Unclaimed Rewards to GSF

hythloda:


waynecollier-da (approved):
Approved with the understanding that we will take the text literally, and the number presented to the Tokenomics will rely on a calculation, performed by IEU (the operator of the GSF SV), of the full total of as-yet unminted Canton Coin escrowed by all SVs currently working toward milestones. The total minted in response to this CIP will allow for the full amount of escrowed rewards to be minted in the future.

PR #74, opened 2025-07-02
CIP-00XX: Enabling Delegated Participation in Liveness Rewards

hythloda:


hythloda:
No SV Sponsor/Endorser to move forward. Please make any needed changes and resubmit. Best process is to send proposal to cip-discuss@lists.sync.global to see what is needed before the GitHub PR. The full process is listed in [CIP0-0000](https://github.com/global-synchronizer-foundation/cips/blob/main/cip-0000/cip-0000.md)

PR #107, opened 2025-09-26
CIP-0082: Establish a 5% Development Fund (Foundation-Governed)

hythloda:


waynecollier-da:
Future implementers and reviewers: Please note that this clarification of the intent of the CIP text:

Where the Specification states, in S2, that "5% of each mint event" will be minted by the Foundation, the intention is that 5% of the total available mint in each round will be minted by the Foundation. This 5% can be implemented as a single single transaction in each round; it should not be implemented as a percentage of each minting transaction.

waynecollier-da (approved):

waynecollier-da:
@hythloda can you update the readme index to say "approved" as well?

PR #162, opened 2026-02-26
CIP-0106.md

pedrodneves:


hythloda:
Seems like this not going forward so closing the PR

PR #168, opened 2026-03-05
contract audit proposal

pixelplex:


meiersi-da (commented):
Thanks a lot. This is a good first stab at the problem.

I think the biggest missing item is support for computing the target vetting state of wallet providers. See my comments.

It might make sense to call out their computation in the intro, and consider the main goal to compute vetting states that allow as many benign apps to be used by users of as many wallet providers as possible, while minimizing the chance of users being exposed to a malicious app.

davidrichards-da (changes requested):
Yuri, I've made a suggestion based on the conversations that I've had with wallet providers.

cip-xxxx-contract-audit/cip-xxxx-contract-audit.md
```diff
@@ -0,0 +1,683 @@
...
+  "schema_version": "1.0",
+  "request_id": "uuid-v4-identifier",
+  "request_timestamp": "2026-02-27T00:00:00Z",
+  "app_provider": {
+    "name": "string",
+    "email": "string",
+    "organization": "string",
+    "repository_url": "https://github.com/org/repo"
+  },
+  "package_information": {
```
jonathanreedmevs, line 252: The CIP appears to strongly bind audits to exact package hashes/package IDs. How are minor or operationally trivial upgrades expected to be handled for live Featured Apps? For example, adding a read-only function would still produce a new package hash and invalidate the existing audit linkage. Is a full re-audit expected for every package change, or should the standard support lighter-weight mechanisms such as delta reviews, audit extensions, or auditor attestations that a previous audit substantially still applies?

cip-xxxx-contract-audit/cip-xxxx-contract-audit.md
```diff
@@ -0,0 +1,683 @@
...
+      "positive_id": "P001",
+      "category": "security|best_practice|design|performance",
+      "title": "string",
+      "description": "string",
+      "components_involved": ["module_name"]
+    }
+  ],
+  "dependency_analysis": {
+    "total_dependencies": 5,
+    "dependencies_reviewed": [
```
jonathanreedmevs, line 382: It may help to clarify what is meant by dependency statuses. In practice, many audit reports only list the dependencies and note observed risks or notable characteristics, rather than making strong security assertions about whether a dependency is “approved.” This distinction seems especially important for community-maintained or widely adopted third-party libraries, where an auditor may be comfortable documenting its usage without wanting to formally endorse the dependency. Some additional guidance or softer terminology here may better reflect an auditor's dependency analysis.

cip-xxxx-contract-audit/cip-xxxx-contract-audit.md
```diff
@@ -0,0 +1,683 @@
...
+          "code_hash": "hex-encoded-hash"
+        }
+      ],
+      "risk_assessment": "string",
+      "remediation": "string",
+      "remediation_status": "open|in_progress|resolved|accepted_risk",
+      "evidence_url": "https://github.com/auditor/audit-reports/blob/main/reports/app-provider/1.0.0/evidence/finding-F001.md"
+    }
+  ],
+  "positives": [
```
jonathanreedmevs, line 371: This feels like a new ask of audit firms that is not normally incorporated into our existing audit processes. I anticipate it being left empty by most firms as it produces overhead to report delivery.

That being said in my experience any positives regarding a specific project would be mentioned in the executive summary of the report.

jonathanreedmevs (commented):
Overall looks great. I think language surrounding things such as security_properties_to_verify needs to be strengthened throughout

cip-xxxx-contract-audit/cip-xxxx-contract-audit.md
jonathanreedmevs, line 429: I'm unclear on what should be contained in `evidence_url` and am worried that it may be redundant on what is already present in `findings.description`

jonathanreedmevs, line 359: This is unclear and could be redundant with what is already present in the report.

cip-xxxx-contract-audit/cip-xxxx-contract-audit.md
```diff
@@ -0,0 +1,806 @@
...
+  Created: 2026-02-27
+  License: CC0-1.0
+* License-Code: CC0-1.0
+* Post-History: 
+* Requires: 
+</pre>
+
+## Abstract
+
+This CIP proposes a standardized protocol for secure interaction between three parties in the Daml application ecosystem: **App Providers**, **Validator Node Providers**, and **Security Auditors**. The standard defines how App Providers publish Daml packages (.dar files) with cryptographically verifiable build metadata, how they request security audits from independent Security Auditors, and how Validator Node Providers can discover and verify audit results to make informed security decisions before integrating third-party packages.
```
meiersi-da, line 20:
```suggestion
This CIP proposes a standardized protocol for secure interaction between three parties in the Daml application ecosystem: **App Providers**, **Validator Node Providers**, and **Security Auditors**. The standard defines how App Providers publish Daml packages (.dar files) with cryptographically verifiable build metadata, how they request security audits from independent Security Auditors, and how Validator Node Providers can discover and verify audit results to make informed security decisions before deploying third-party packages on their nodes.
```

cip-xxxx-contract-audit/cip-xxxx-contract-audit.md
```diff
@@ -0,0 +1,806 @@
...
+### Overview
+
+#### Parties
+
+- **App Provider**: An organization or individual developing Daml applications and publishing Daml packages (.dar files)
+- **Validator Node Provider**: A service provider providing users with access to a Canton Network validator node hosting their parties.
+- **Security Auditor**: An independent security firm or team validating the security properties of Daml packages
+
+#### Problem Statement
+
```
meiersi-da, line 41: To me the core problem is how to ensure that:
1. app users can safely use all the applications they would like to use
2. app providers can easily roll-out upgrades to their applications

We can the observe that validator node operators take the role of shielding app users from the technical challenge of determining what app upgrades to accept when. And for validator node operators to do that efficiently and effectively, they need information from both app providers and security auditors.

So with that perspective, I'd suggest:
1. add "app user" to the parties list above
2. review the list of properties below, as to whether they are really required for validator node operators to make good decisions as to what apps and app upgrades to deploy on their nodes.

cip-xxxx-contract-audit/cip-xxxx-contract-audit.md
```diff
@@ -0,0 +1,806 @@
...
+1. Discover what security validations have been performed on third-party packages
+2. Verify the integrity and authenticity of security audit reports
+3. Make informed decisions about package integration based on transparent audit results
+4. Establish trust without performing redundant security validations
+5. Understand version compatibility between composed applications
+
+App Providers benefit from:
+1. Clear specification of what information to provide for auditing
+2. Decentralized publication of audit results
+3. Reuse of existing audits across multiple Validator Node Providers
```
meiersi-da, line 52: I'm missing the aspect of low friction roll-out of apps and their upgrades to a wide user-base here. To me that's the most important aspect.

cip-xxxx-contract-audit/cip-xxxx-contract-audit.md
```diff
@@ -0,0 +1,806 @@
...
+
+#### App Provider Git Repository Structure
+
+Each App Provider maintains a public or private Git repository with the following structure:
+
+```
+app-provider-repo/
+├── README.md
+├── apps/
+│   ├── canton-coin/
```
meiersi-da, line 64: This seems like an example, but above it reads like this is mandatory structure.

cip-xxxx-contract-audit/cip-xxxx-contract-audit.md
```diff
@@ -0,0 +1,806 @@
...
+  "release_date": "2026-02-27T00:00:00Z",
+  "description": "string",
+  "publisher": {
+    "name": "string",
+    "email": "string",
+    "organization": "string",
+    "public_key_id": "string",
+    "repository_url": "https://github.com/org/repo"
+  },
+  "package_hash": {
```
meiersi-da, line 114: is this the hash of the .dar file or the package-id of the .dalf files in there?

cip-xxxx-contract-audit/cip-xxxx-contract-audit.md
```diff
@@ -0,0 +1,806 @@
...
+  "build_reproducibility": {
+    "daml_sdk_version": "3.3.0",
+    "ghc_version": "9.2.4",
+    "build_timestamp": "2026-02-27T00:00:00Z",
+    "build_parameters": {
+      "optimization_level": "release",
+      "additional_flags": ["--ghc-option=-O2"]
+    },
+    "build_instructions_url": "https://github.com/org/repo/blob/main/BUILD.md#v1.0.0"
+  },
```
meiersi-da, line 139: Ideally, this is encoded in the .dar file itself.

@paulbrauner-da: what's the status of being able to recompile a .dar to check its source to content mapping?

cip-xxxx-contract-audit/cip-xxxx-contract-audit.md
```diff
@@ -0,0 +1,806 @@
...
+  "security_properties": {
+    "daml_stdlib_version": "3.3.0",
+    "notable_modules": [
+      {
+        "module_name": "App.Security.Authorization",
+        "description": "Custom authorization logic"
+      }
+    ]
```
meiersi-da, line 147: Why do we need this section?

cip-xxxx-contract-audit/cip-xxxx-contract-audit.md
```diff
@@ -0,0 +1,806 @@
...
+  },
+  "licenses": {
+    "package_license": "Apache-2.0",
+    "dependency_licenses": [
+      {
+        "package_name": "string",
+        "license": "Apache-2.0"
+      }
+    ]
+  },
```
meiersi-da, line 157: This is part of the .dar AFAIK. Do we need it here?

cip-xxxx-contract-audit/cip-xxxx-contract-audit.md
```diff
@@ -0,0 +1,806 @@
...
+  }
+}
+```
+
+##### Security Properties Definitions
+
+- **`daml_stdlib_version`**: The version of the Daml Standard Library used to compile the package.
+- **`notable_modules`**: A list of modules that are particularly important for security review (e.g., core business logic, authorization modules).
+
+#### Build Configuration Format
```
meiersi-da, line 170: Let's see that we can move this out of this CIP into a feature of .dar files themselves.

cip-xxxx-contract-audit/cip-xxxx-contract-audit.md
```diff
@@ -0,0 +1,806 @@
...
+  "package_hash": {
+    "algorithm": "sha256",
+    "value": "c208d7ead1e4e9b610fc2054d0bf00716144ad444011bce0b02dcd6cd0cb8a23"
+  },
+  "audit_results": [
+    {
+      "auditor_name": "Security Auditor Company",
+      "audit_date": "2026-02-20T00:00:00Z",
+      "audit_report_url": "https://github.com/auditor/audit-reports/blob/main/reports/app-provider-name/1.0.0/audit-report.json",
+      "audit_status": "passed",
```
meiersi-da, line 227: Might come later: I think we need a basic security definition here: the key aspect is that a package does not abuse the user's authority in the contracts to which the user delegated the authority.

cip-xxxx-contract-audit/cip-xxxx-contract-audit.md
```diff
@@ -0,0 +1,806 @@
...
+    ],
+    "unvet": [
+      {
+        "package_id": "splice-amulet",
+        "package_version": "0.1.15",
+        "package_hash": {
+          "algorithm": "sha256",
+          "value": "..."
+        }
+      }
```
meiersi-da, line 272: The common case I see is to: unvet all packages with the "splice-amulet" name, and then vet the specific ones that should be active.

cip-xxxx-contract-audit/cip-xxxx-contract-audit.md
```diff
@@ -0,0 +1,806 @@
...
+          "value": "c208d7ead1e4e9b610fc2054d0bf00716144ad444011bce0b02dcd6cd0cb8a23"
+        }
+      },
+      {
+        "package_id": "splice-util",
+        "package_version": "0.1.5",
+        "package_hash": {
+          "algorithm": "sha256",
+          "value": "5a58024e2cc488ca9e0c952ec7ef41da3a1ed0a78ba23bacd819e5b30afb5546"
+        }
```
meiersi-da, line 261: let's add an example with effective dating

cip-xxxx-contract-audit/cip-xxxx-contract-audit.md
meiersi-da: Right now, a basic audit needs to be done on every change to a package, as each of them could in principle introduce a choice that abuses an existing delegation of a user's authority.

I would expect that audit though to be done in a delta form of new sources against the previously audited sources.

cip-xxxx-contract-audit/cip-xxxx-contract-audit.md
```diff
@@ -0,0 +1,806 @@
...
+  "signature": {
+    "algorithm": "rsa-sha256|ecdsa-sha256",
+    "public_key_id": "string",
+    "signature_value": "hex-encoded-signature"
+  },
+  "additional_resources": {
+    "detailed_findings_url": "https://github.com/auditor/audit-reports/blob/main/reports/app-provider/1.0.0/detailed-findings.md",
+    "evidence_artifacts_url": "https://github.com/auditor/audit-reports/tree/main/reports/app-provider/1.0.0/evidence"
+  }
+}
```
meiersi-da, line 483: This is a lot of data. I'm not convinced that we need all of this for the purpose of making it easy for apps to get deployed on validator nodes.

I'd suggest we focus on the minimal properties that make it safe for a user to engage with an app.

cip-xxxx-contract-audit/cip-xxxx-contract-audit.md
```diff
@@ -0,0 +1,806 @@
...
+```
+
+#### Security Auditor Index File
+
+Each auditor maintains an `index.json` for discoverability:
```
meiersi-da, line 488: Can you explain how this would be used by validator node operators to make their decision on whether to deploy an app (ugprade)?

cip-xxxx-contract-audit/cip-xxxx-contract-audit.md
```diff
@@ -0,0 +1,806 @@
...
+
+#### Validator Node Provider Verification Procedure
+
+Validator Node Providers implement the following process before integrating a third-party package:
+
+```json
+{
+  "schema_version": "1.0",
+  "verification_process": {
+    "step_1_locate_package": {
```
meiersi-da, line 536: Why are these steps described as JSON? I'd have expected them to be addressed to humans...

cip-xxxx-contract-audit/cip-xxxx-contract-audit.md
```diff
@@ -0,0 +1,806 @@
...
+        "combine_internal_and_external"
+      ]
+    }
+  }
+}
+```
+
+#### Validator Node Provider Integration Decision Document
+
+Validator Node Providers maintain records of their integration decisions:
```
meiersi-da, line 580: Why? They may do so, but do they need to do so?

cip-xxxx-contract-audit/cip-xxxx-contract-audit.md
```diff
@@ -0,0 +1,806 @@
...
+
+1. Each Validator Node Provider must either audit packages independently (expensive and duplicative) or trust App Providers without verification (risky)
+2. Security audits lack standardization, making it difficult to compare results across different auditors
+3. There is no transparent way for audits to be discovered and verified
+4. Audit results cannot be reused across multiple Validator Node Providers, leading to wasted effort
+
+This CIP solves these problems by establishing a transparent, standardized ecosystem where:
+
+- App Providers clearly communicate what they're publishing and how it was built
+- Security Auditors can publish results in verifiable repositories
```
meiersi-da, line 673: In what sense are the repositories "verifiable"?

cip-xxxx-contract-audit/cip-xxxx-contract-audit.md
```diff
@@ -0,0 +1,806 @@
...
+- Validator Node Providers can discover and verify audits without conducting redundant validations
+- All parties operate with cryptographic verification and transparency
```
meiersi-da, line 675: I see hashes being used, but no signatures. Do we need code-signing to guard against compromise of the GH repos?

cip-xxxx-contract-audit/cip-xxxx-contract-audit.md
```diff
@@ -0,0 +1,683 @@
...
+3. There is no transparent way for audits to be discovered and verified
+4. Audit results cannot be reused across multiple WalletProviders, leading to wasted effort
+
+This CIP solves these problems by establishing a transparent, standardized ecosystem where:
+
+- AppProviders clearly communicate what they're publishing and how it was built
+- SecurityAuditors can publish results in verifiable repositories
+- WalletProviders can discover and verify audits without conducting redundant validations
+- All parties operate with cryptographic verification and transparency
+
```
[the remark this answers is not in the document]
meiersi-da: @davidrichards-da: note that version hell should go away with packages moving away from depending on other app's templates directly and only compose via interfaces.

cip-xxxx-contract-audit/cip-xxxx-contract-audit.md
```diff
@@ -0,0 +1,806 @@
...
+Providers are concerned about the operational effort required to load DARs onto validators. They’d like this as automated as possible.
+
+## Rationale
+
+### Design Decisions
+
+#### Why Git-based Publishing
+
+Git provides:
+- **Cryptographic verification** via commit signatures
```
meiersi-da, line 699: Who verifies these signatures and where does this verification happen?

cip-xxxx-contract-audit/cip-xxxx-contract-audit.md
```diff
@@ -0,0 +1,806 @@
...
+- Validator Node Providers can adopt verification procedures at their own pace
+
+Existing packages without audits remain usable; Validator Node Providers simply cannot leverage audit information for integration decisions.
+
+## Implementation
+
+An implementation of this standard should include:
+
+1. **App Provider tooling**:
+   - CLI tool to generate `metadata.json` and `build-config.json` from .dar files
```
meiersi-da, line 782:
```suggestion
```
As written above, I believe build reproducibility is better solved by changing `dpm` to check the sources of a .dar matching its contents.

So I'm not sure about this tool being built.

However, a tool that does seem useful is one that generates a target vetting state from a set of .dars

cip-xxxx-contract-audit/cip-xxxx-contract-audit.md
```diff
@@ -0,0 +1,806 @@
...
+
+2. **Security Auditor tooling**:
+   - Templates for `audit-report.json` generation
+   - Signature generation and verification utilities
+   - Batch reporting tools for multiple packages
+
+3. **Validator Node Provider tooling**:
+   - Library to fetch and verify audit reports
+   - Decision engine implementing verification procedures
+   - UI components for displaying audit results to users
```
meiersi-da, line 794: Why do they need to publish this to users?

cip-xxxx-contract-audit/cip-xxxx-contract-audit.md
```diff
@@ -0,0 +1,806 @@
...
+   - Documentation for setting up audit requests
+
+2. **Security Auditor tooling**:
+   - Templates for `audit-report.json` generation
+   - Signature generation and verification utilities
+   - Batch reporting tools for multiple packages
+
+3. **Validator Node Provider tooling**:
+   - Library to fetch and verify audit reports
+   - Decision engine implementing verification procedures
```
meiersi-da, line 793: Is this the script that computes the target vetting state for their validator nodes from a spec of the apps that they would like to deploy and the security audit reports to use?

meiersi-da (commented):
Thanks for this second iteration. I've focused my review on the CIP draft.

Overall I think it contains most of the elements required for a first solution to the problem. However, it also contains quite a few pieces whose role in solving the key problem (see my comments) is not (yet) clear to me.

For standards, I'd dare say that "less is more". So I'd appreciate if you could have a look at my comments and the CIP text itself, and see what content can be culled without impacting the CIP's fitness for purpose.

Happy to make another pass once you've done so.

TracyLi025:
Hi, we at CertiK would also like to support this proposal. There are a few items that we think are worth discussing. 

### 1. What auditors can and cannot sign

This is the most important item for us, and we expect it to apply to every report issued under this standard.

In practice, the client frequently requires the auditing firm to keep the report confidential or to redact portions of it, so some of the information the standard expects may not be available to the public.

Our suggestion is to separate the attestation from the report. The auditor publishes a small, signed attestation carrying only what a validator actually needs:

- `package_id`
- `dar_sha256`
- `scope[]` and `out_of_scope[]`
- `trust_assumptions[]` made during the audit
- `status`, together with a `severity_scale` identifier
- `daml_sdk_version` at the time of the audit
- `dependencies` / standards relied upon
- `audit_completion_date`
- `Full_report_hash`, binds the attestation to a specific report without disclosing it
- `disclosure_policy`: `private`, `embargoed_until: <date>`, or `public`
- `signature`

The full report then remains under whatever terms the App Provider and the auditor agree on. Because the attestation carries the report hash, the report stays verifiable whenever it is eventually disclosed, nothing is lost on the integrity side.

### 2. Identity binding and signature mechanics

As currently specified, the signature scheme is effectively arbitrary: a signature verifies against a public key, but nothing binds that key to the firm it claims to represent. The standard should define a way to bind the published signature to the auditing company. For example, keys distributed through a verifiable channel such as the auditor's official domain or an ecosystem-maintained registry.

### 3. Comparability across auditors

Severity ratings are not comparable across firms today. With no declared scale, `severity: "high"` has no fixed meaning, and App Providers will naturally route work to whichever firm rates most leniently. The standard would end up producing a race to the bottom instead of preventing one.

We have two suggestions.

**Mechanically**, make `severity_scale` and a methodology reference mandatory alongside every severity value, so ratings always carry their provenance. An optional CVSS vector could be included where it applies.

**Substantively**, there is no Daml or Canton-specific severity taxonomy yet, and the EVM-derived categories most of the industry uses (reentrancy, gas, integer overflow) map poorly onto this platform. The actual risk surface here is signatory and controller authorization logic, party confidentiality via disclosure, contract key uniqueness, and upgrade compatibility.

For reference, the classification we currently use:

| Severity | Description |
| --- | --- |
| **Critical** | Critical risks are those that impact the safe functioning of a platform and must be addressed before launch. Users should not invest in any project with outstanding critical risks. |
| **Major** | Major risks may include logical errors that, under specific circumstances, could result in fund losses or loss of project control. |
| **Medium** | Medium risks may not pose a direct risk to users' funds, but they can affect the overall functioning of a platform. |
| **Minor** | Minor risks can be any of the above, but on a smaller scale. They generally do not compromise the overall integrity of the project, but they may be less efficient than other solutions. |
| **Informational** | Informational errors are often recommendations to improve the style of the code or certain operations to fall within industry best practices. They usually do not affect the overall functioning of the code. |

### 4. Clarification on the `expiration_date` field

An audit attests to a specific code state and should not expire merely because time has passed. The only event that should invalidate it is a change to the code after the audit. We suggest reframing this field so it deprecates an audit once the audited code is updated, rather than imposing a calendar-based expiry.

PR #169, opened 2026-03-05
profile credentials proposal

pixelplex:


meiersi-da (commented):
Thanks a lot for the nice work. Neat and tight. 💪

cip-xxxx-profile-credentials/cip-xxxx-profile-credentials.md
```diff
@@ -0,0 +1,317 @@
...
+Status: Draft
+Title: Canton Network Party Profile Credentials
+Type: Standards Track
+---
+
+# Abstract
+
+This CIP standardizes a portable representation of **party profile
+metadata** for user‑interface rendering on the Canton Network based on
+the **Canton Network Credentials Standard**.
```
leonidr-c7: What is this a reference to ?

cip-xxxx-profile-credentials/cip-xxxx-profile-credentials.md
```diff
@@ -0,0 +1,317 @@
...
+
+## `cip-<nr>/website`
+
+Website reference associated with the party.
+
+Applications:
+
+-   MUST treat the value as informational metadata
+-   MUST NOT treat it as an authoritative identifier
+-   SHOULD expect an `https://` URL
```
leonidr-c7: Should we tighten the language here and require a URL that conforms to a spec?

cip-xxxx-profile-credentials/cip-xxxx-profile-credentials.md
```diff
@@ -0,0 +1,317 @@
...
+-   MAY display invalid URLs as plain text
+-   MUST NOT treat invalid values as trusted links
+
+## `cip-<nr>/email`
+
+Informational contact email.
```
leonidr-c7, line 140: Similar to the above, and applies to other claims below so I won’t repeat, should we require that these values conform to their specifications?

cip-xxxx-profile-credentials/cip-xxxx-profile-credentials.md
```diff
@@ -0,0 +1,317 @@
...
+A source is defined as:
+
+`(registry, issuer)`
+
+Issuer may be:
+
+-   a specific issuer party ID
+-   `self` (meaning issuer = holder party)
+
+## Resolution Algorithm
```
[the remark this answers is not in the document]
leonidr-c7: I’d second @meiersi-da request to split out the resolution algorithm to a separate CIP.

cip-xxxx-profile-credentials/cip-xxxx-profile-credentials.md
```diff
@@ -0,0 +1,317 @@
...
+
+Applications:
+
+-   MUST treat the value as informational metadata
+-   MAY display the value as provided
+
+# Party Profile Resolution
+
+Applications may obtain profile credentials for a party **P** from
+multiple registries and issuers.
```
leonidr-c7: Should we make Daml specific claims about an `issuer` or `registrar`?

Specifically, I would expect that the issuer gave authorization, is a signatory, on a contract that makes a claim about a user. Furthermore, it would be valuable for there to be more than one issuer; that would allow us to make stronger claims. Ex.

A token where both `leonid::1220…` and `c7trust::1220…` are issuers (signatories) on a claim that `leo@c7.digital` is my email address, that carries more weight then one where it is just `leonid:1220…` 😄

cip-xxxx-profile-credentials/cip-xxxx-profile-credentials.md
```diff
@@ -0,0 +1,317 @@
...
+## Why Namespaced Claim Keys
+
+Using namespaced keys `cip-<nr>/*`:
+
+-   minimizes protocol changes
+-   preserves compatibility with existing credential infrastructure
+-   enables efficient querying
+-   allows extensibility
+
+Future profile attributes can be added under the same namespace.
```
leonidr-c7: But wouldn’t future attributes have to come from a different CIP?

leonidr-c7 (commented):
Feedback

cip-xxxx-profile-credentials/cip-xxxx-profile-credentials.md
vkokosh: Since the standard is still in draft, I didn’t think an explicit reference was necessary at this stage, but I can add it.

cip-xxxx-profile-credentials/cip-xxxx-profile-credentials.md
vkokosh: As far as I remember, Simon and I discussed this as a separate standard, but I could be wrong. However, I will clarify the wording and add more strict guidelines for verification.

cip-xxxx-profile-credentials/cip-xxxx-profile-credentials.md
vkokosh: As Simon and I discussed, new well-known attributes don’t have to mean a brand-new CIP every time - they can be added via amendments to this CIP. The namespace is mainly to keep clients forward-compatible with unknown keys.

cip-xxxx-profile-credentials/cip-xxxx-profile-credentials.md
leonidr-c7: Thanks @vkokosh I found Simon’s gdoc for this.

cip-xxxx-profile-credentials/cip-xxxx-profile-credentials.md
```diff
@@ -0,0 +1,317 @@
...
+## `cip-<nr>/github`
+
+GitHub username or organization.
+
+Applications:
+
+-   MUST treat the value as informational metadata
+-   MAY link to a GitHub profile URL
+
+## `cip-<nr>/discord`
```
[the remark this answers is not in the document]
meiersi-da: hmm... the convention to use `cip-<nr>/` prefixes is proving quite valuable, as it makes it very easy to go from data observed in prod to the explanation of the data. I'd rather have your fields be prefixed by the `cip-<nr>` of your CIP.

cip-xxxx-profile-credentials/cip-xxxx-profile-credentials.md
```diff
@@ -0,0 +1,339 @@
+---
+Author: PixelPlex Inc.
+CIP: TBD
+Created: 2026-01-27
+License: CC0-1.0
+Status: Draft
+Title: Canton Network Party Profile Credentials
+Type: Standards Track
+---
```
meiersi-da: Consider adjusting to the standard defined in the repo: https://github.com/canton-foundation/cips/blob/main/cip-XXXX/cip-XXXX.md

cip-xxxx-profile-credentials/cip-xxxx-profile-credentials.md
```diff
@@ -0,0 +1,339 @@
...
+Author: PixelPlex Inc.
+CIP: TBD
+Created: 2026-01-27
+License: CC0-1.0
+Status: Draft
+Title: Canton Network Party Profile Credentials
```
meiersi-da:
```suggestion
Title: Canton Network Party Profiles
```

cip-xxxx-profile-credentials/cip-xxxx-profile-credentials.md
```diff
@@ -0,0 +1,339 @@
...
+It defines:
+
+-   A set of **standard claim keys** for common party metadata such as
+    display name, avatar, website, and optional contact/social
+    information.
+-   A clear **application-side interpretation model** for rendering these
+    claims consistently in user interfaces.
+
+The profile claims defined by this CIP are **informational only** and
+**MUST NOT** be interpreted as verified identity attributes or used for
```
meiersi-da:
```suggestion
**SHOULD NOT** be interpreted as verified identity attributes or used for
```

SHOULD as they may be used for these purposes if extra information about the issuer of the credential is known.

meiersi-da: Hmm... I see that you restrict the profile below to self-published credentials. It remember we had some conversation about this; and I might have pushed in this direction. However, I think this might be too limiting in the following sense: we would like to create space for app providers to publish metadata about parties they interact with. We should do so in a form that (a) makes it easy for app providers to start doing so at their own pace and (b) allows apps to converge on common metadata keys.

Given this context, I'd propose that we extend the display rules such that the profile consist of sections, where the first section contains the self-published information under well-known keys, and then there's one section per issuer known to the displaying app (i.e. whitelisted by them), and then there follow sections for data from unknown issuers. For each of the sections, they first show the known claims, and afterwards the unknown claims.

wdyt?

cip-xxxx-profile-credentials/cip-xxxx-profile-credentials.md
```diff
@@ -0,0 +1,339 @@
...
+`cip-<nr>/`
+
+Where `<nr>` will be replaced by the assigned CIP number.
+
+# Party Profile Claim Keys
+
+For social fields, this CIP uses grouped claim keys (for example,
+`cip-<nr>/social:discord`) to align naming conventions with
+[CPRP (CIP PR #171)](https://github.com/canton-foundation/cips/pull/171)-style
+field organization while keeping this CIP namespace prefix.
```
meiersi-da:
```suggestion
```

This was removed from the CPRP PR.

cip-xxxx-profile-credentials/cip-xxxx-profile-credentials.md
```diff
@@ -0,0 +1,339 @@
...
+properties are required.
+
+Applications MUST use only active credential records. A credential MUST
+be ignored when the lookup time is before `validFrom`, after `validUntil`,
+or after the registry record's `expiresAt`, when the respective field is
+present. Archived credentials MUST be ignored. The precise time-boundary
+semantics MUST follow the Canton Network Credentials Standard.
+
+Applications SHOULD query the DSO Credential Registry by default and MAY
+query additional registries according to an explicit local policy. When
```
meiersi-da:
```suggestion
query additional registries according to an explicit application specific policy. When
```

cip-xxxx-profile-credentials/cip-xxxx-profile-credentials.md
```diff
@@ -0,0 +1,339 @@
...
+
+Applications:
+
+-   MUST treat the value as a display string
+-   MUST NOT treat it as an identifier
+-   SHOULD render the value as-is after applying UI escaping
+-   MAY gracefully truncate the rendered value
+
+The kebab-case `display-name` property follows the naming convention used
+by the Canton Network Credentials Standard.
```
meiersi-da: Move into preamble

cip-xxxx-profile-credentials/cip-xxxx-profile-credentials.md
```diff
@@ -0,0 +1,339 @@
...
+identity, or prove ownership of an external account or resource.
+
+Profile metadata is expressed as **credential claims** under a dedicated
+namespace:
+
+`cip-<nr>/`
+
+Where `<nr>` will be replaced by the assigned CIP number.
+
+# Party Profile Claim Keys
```
meiersi-da:
```suggestion
## Party Profile Claim Keys
```

cip-xxxx-profile-credentials/cip-xxxx-profile-credentials.md
```diff
@@ -0,0 +1,339 @@
...
+Credential producers:
+
+-   MUST encode a value conforming to Discord username or user ID format rules
+
+Applications:
+
+-   MUST treat the value as informational metadata
+-   MAY display the value as provided
+
+# Party Profile Resolution
```
meiersi-da:
```suggestion
## Party Profile Resolution
```

cip-xxxx-profile-credentials/cip-xxxx-profile-credentials.md
```diff
@@ -0,0 +1,339 @@
...
+-   claim interpretation semantics
+-   credential filtering
+-   deterministic per-property resolution
+
+but leaves configurable:
+
+-   selection of additional registries
+-   registry priority ordering
+
+# Examples
```
meiersi-da: Move examples below "Rationale" to satisfy CIP format: https://github.com/canton-foundation/cips/blob/main/cip-XXXX/cip-XXXX.md

cip-xxxx-profile-credentials/cip-xxxx-profile-credentials.md
```diff
@@ -0,0 +1,339 @@
...
+by the Canton Network Credentials Standard.
+
+## `cip-<nr>/avatar`
+
+Avatar reference for UI rendering.
+
+Credential producers:
+
+-   MUST encode a URI conforming to RFC 3986
+-   MUST use an `https://` URL or `ipfs://` URI
```
meiersi-da: Compared to ENS there's no `data` support: https://docs.ens.domains/web/avatars/

Consider explaining why not in rationale. I suspect there's not enough space is there?

cip-xxxx-profile-credentials/cip-xxxx-profile-credentials.md
```diff
@@ -0,0 +1,339 @@
...
+-   MUST NOT treat it as an identifier
+-   SHOULD render the value as-is after applying UI escaping
+-   MAY gracefully truncate the rendered value
+
+The kebab-case `display-name` property follows the naming convention used
+by the Canton Network Credentials Standard.
+
+## `cip-<nr>/avatar`
+
+Avatar reference for UI rendering.
```
meiersi-da:
```suggestion
Avatar reference for UI rendering as a user profile image.
```

cip-xxxx-profile-credentials/cip-xxxx-profile-credentials.md
```diff
@@ -0,0 +1,339 @@
...
+-   MUST validate the URI before using it
+-   MAY ignore values that are not valid or supported URIs
+
+## `cip-<nr>/website`
+
+Website reference associated with the party.
+
+Credential producers:
+
+-   MUST encode an absolute `https://` URL conforming to RFC 3986
```
meiersi-da: These are self-published credentials. There's no validation going on when producing them. How do you expect this guidance to be interpreted in practice?

Consider using the more general term "Publishers", which includes self-publication and other issuers; and also call out that applications MAY ignore values that do not conform to the spec.

cip-xxxx-profile-credentials/cip-xxxx-profile-credentials.md
meiersi-da: There are a bunch of keys missing compared to what ENS supports: https://docs.ens.domains/ensip/5#global-keys

Consider adding them. Also consider adding a "Rationale > Relation to ENS" section that explains that ENS is used as inspiration, and ENSIP-5 is used as the basis for the keys defined here with the exception of <list all exceptions>. Consider minimizing the exceptions to enable easy interoperability should we want to enable that in the future.

cip-xxxx-profile-credentials/cip-xxxx-profile-credentials.md
```diff
@@ -0,0 +1,339 @@
...
+or after the registry record's `expiresAt`, when the respective field is
+present. Archived credentials MUST be ignored. The precise time-boundary
+semantics MUST follow the Canton Network Credentials Standard.
+
+Applications SHOULD query the DSO Credential Registry by default and MAY
+query additional registries according to an explicit local policy. When
+multiple registries are queried, applications MUST apply a deterministic
+registry-priority order.
+
+Profile properties MUST be resolved independently. After the filtering
```
meiersi-da: Some profile properties like `email` make sense to be set multiple times or not? One can probably argue both ways. It would though be good to give a clear argument in the rationale section.

Also consider adding subsection titles or section headings that summarize what's being specified. Here for example "Claim deduplication:"

cip-xxxx-profile-credentials/cip-xxxx-profile-credentials.md
```diff
@@ -0,0 +1,339 @@
...
+above, a value from a higher-priority registry takes precedence. Within
+one registry, the credential with the latest registry record time takes
+precedence. If record times are equal, the contract ID MUST be used as the
+deterministic tie-breaker in the order defined by the Canton Network
+Credentials Standard.
+
+The resolution/composition flow defined in
+[CPRP (CIP PR #171)](https://github.com/canton-foundation/cips/pull/171)
+MAY be used as a higher-level policy only if it preserves the mandatory
+lookup, filtering, and deterministic resolution requirements above.
```
meiersi-da:
```suggestion
```

This one is gone.

cip-xxxx-profile-credentials/cip-xxxx-profile-credentials.md
```diff
@@ -0,0 +1,339 @@
...
+but leaves configurable:
+
+-   selection of additional registries
+-   registry priority ordering
+
+# Examples
+
+## Example 1 --- Basic Profile Claims
+
+A self-published credential, for which `issuer = holder`, contains:
```
meiersi-da, line 523:
```suggestion
A self-published credential, for which `issuer = holder = <pixelplex-party>`, contains:
```

Question: what kind of PixelPlex party would have the profile you show here? I'm asking as the profile is kind of very top-level and doesn't leave a lot of space for other parties owned by PixelPlex.

Consider giving an example of a party representing the Console wallet, your explorer, or your public facing treasury (assuming you'd want to have a profile for that). Also consider giving an example for the profile of a Console wallet user.

meiersi-da (commented):
Very nice work! Wow... I'm sooo looking forward to seeing the first real profile page in CC View! 

I think this is already a great basis, and I really like the direction in which this is going. If we manage to find a good answer to the issuer-published profile information and apply the remaining polish then we're done with this.

(And yes: I'll see that I get the CN Credential Standard done as well and kick-off it's implementation.)

cip-xxxx-profile-credentials/cip-xxxx-profile-credentials.md
pixelplex: @meiersi-da i was thinking about example with our console wallet for instance, where 
issuer = holder = consolewallet.cns (party pixelplex-mainnet-3::12209a8ec282ec94bc941172f3052221413031f27394685693e2d83e4bfab6423ac6)

 so the wallet users eventually can own a record  something like `alice.consolewallet.cns`

cip-xxxx-profile-credentials/cip-xxxx-profile-credentials.md
VladislavKokosh: For now I’m going with a simple rule in the spec: one value per well-known key inside a single issuer section (dedup / last-write-wins). Multiple values still show up across issuers, because each issuer is its own section. So email from self-published and email from a whitelisted app can both appear; two emails from the same issuer collapse to one.
Сlaims are a map, and a section stays predictable if each key has at most one value. If someone needs a second email under the same issuer, that can be a later key/amendment rather than making every field multi-value now.
I’ll put this in the rationale under Claim deduplication. If you’d rather allow multiple emails from the same issuer, we can change it.

PR #171, opened 2026-03-06
CIP: Canton Party Name Resolution Standard

pdomenighetti:
PR from Freename AG


cprp-resolution.md
```diff
@@ -0,0 +1,280 @@
...
+### Resolver Interface
+
+Any identity provider can become a CPRP resolver by implementing the following interface:
+
+| Method | Input | Output | Purpose |
+|--------|-------|--------|---------|
+| `resolve` | namespace, name | `ResolutionResult` | Forward lookup: name → Party ID + metadata |
+| `reverseResolve` | partyId | `ReverseResolutionResult` | Reverse lookup: Party ID → registered names |
+| `resolveMulti` | queries[] | `ResolutionResult[]` | Batch resolution (latency optimization) |
+| `changelog` | since timestamp | `ChangelogEntry[]` | Credential changes since a given time |
```
meiersi-da: Are these HTTP methods? Can you add references to their specification?

cprp-resolution.md
```diff
@@ -0,0 +1,280 @@
...
+
+Each resolver publishes backing credentials via the CN Credentials Daml interface and reports its type:
+
+| Resolver Type | Registration | Trust Tier | Example |
+|--------------|-------------|-----------|---------|
+| Featured | SV governance vote (on-ledger `ResolverFeaturedStatus`) | T3 | Freename, 7Trust |
+| Permissionless | No registration required; implements the interface | T4 | Any third-party identity service |
+| Address Book | Local to the application; no network registration | App-defined | Citadel internal directory |
```
meiersi-da: The `T3`, `T4` qualifiers introduce a bit of a cyclic reference. I wonder whether you can either write this CIP completely w/o such a forward reference on another CIP; or whether it is really one CIP being proposed, that just uses an extended format for its specification to make it more readable.

meiersi-da (commented):
Very nice work @pdomenighetti ! Thank you 🙏 

I believe this has the right ingredients, and will result in a great improvement for the ecosystem after a few rounds of polishing. See my comments to drive a first iteration.

Some of the comments hint at scope that we might be able to remove: e.g, encrypted fields. In my experience, keeping scope tight is very valuable to focus our limited attention on the pieces with the highest ROI; and deliver the overall improvement in a timely fashion.

cprp-resolution.md
```diff
@@ -0,0 +1,280 @@
...
+
+| Component | Purpose | Example Values |
+|-----------|---------|---------------|
+| `network` | Prevents cross-environment confusion | `mainnet`, `testnet`, `devnet` |
+| `resolver` | Identity source that backs the name | `dns`, `vlei`, `freename`, `self` |
+| `namespace` | Registrant's domain or organizational scope | `goldmansachs.com`, `acme-bank.canton` |
+| `n` | Specific name within the namespace | `default`, `treasury`, `trading-desk-3` |
+
+Examples:
+- `mainnet/dns:goldmansachs.com:default`
```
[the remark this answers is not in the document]
meiersi-da: consider avoiding the `mainnet/` prefix, as parties are not bound to networks. You might use `party:` without a qualifier or use `protocol/party`. Or perhaps avoid even stating that as party and just use the Canton concept: `uid:`, which stands for Universal Identifier, and is cryptic enough so that people don't attach too many unwarranted concepts to it.

cprp-resolution.md
```diff
@@ -0,0 +1,307 @@
...
+1. Groups results by `party_id`
+2. For same-resolver duplicates: selects by Ledger Effective Time (last-write-wins, per CNS 2.0)
+3. For cross-resolver duplicates: merges metadata, selects display name by weight
+4. For conflicting `party_id` values: triggers collision handling per the strategy's `collision_policy`
+5. Merges credential arrays, endpoint maps, and profile claims across all sources
+6. Records per-claim provenance (`claim_sources`): for each metadata key, tracks which resolver and issuer contributed the value — enabling institutional audit trails
+
+### Profile Rendering Guidelines
+
+Profile claims (`cns-2.0/name`, `cns-2.0/avatar`, `cns-2.0/email`, `cns-2.0/website`) are informational only and MUST NOT be interpreted as verified identity attributes. Verification status is determined exclusively by CIP-YYYY's trust evaluation, not by profile content. Display names SHOULD be ≤64 Unicode characters. Avatars SHOULD be `https://` URLs; applications MAY additionally support `ipfs://` URIs.
```
meiersi-da: Unicode has a lot of typo-squatting space, and weird behavior in undefined ranges. Do you know what the latest guidance is for allowing Unicode in names?

CPRP-spec.md
```diff
@@ -0,0 +1,3310 @@
...
+
+Where <network> is one of:
+  mainnet   — Canton MainNet (production)
+  testnet   — Canton TestNet
+  devnet    — Canton DevNet
+```
+
+Rules:
+
+1. All on-ledger credentials MUST include the `cprp/network` claim key set to the network identifier.
```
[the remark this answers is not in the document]
meiersi-da: Would you be opposed to adding a preamble the `cprp/` is the temporary prefix until a CIP number is assigned?

meiersi-da (commented):
Thanks a lot for the extensive detail in the provided CIPs.

I've done a pass over CPRP-resolution in particular. The ideas look comprehensive, but I have not yet had enough time to engage with them to fully understand how they would work in practice and where their snags are. Looking forward to doing so once the time is available.

leonidr-c7:
### Following up on @dave-axymos and @meiersi-da 

I do not understand why we need a uniform resolution proposal. These resolution algorithms, will fundamentally rely on who is issuing a given token and what the consumers of those tokens want with respect to an acceptable trust model. It seems very early to try and standardize how an entire network  _SHOULD_ resolve claims.


### Wrt to the genera import question.

@meiersi-da I disagree with

> For the "import" problem, we can define standard verification methods, 

There are very different claims of

1. self-attesting to a DNS entry, 
2. a multi signed DNS attestation after a challenge/response.
3. a self-attesting LEI entry.
4. a self-attesting vLEI entry.
5. a multi-sign vLEI entry because an attesting party runs KERI verification
6. a `Party` that is backed by keys on the KERI node, ie a native IDP integration.


### Suggestion.

Can we have a standalone CIP that focuses on creating  `.canton` , I like the idea’s @dave-axymos is describing with XNS and think some things from this one could be repurposed.

zhihezhang2:
Thanks @pdomenighetti for the very thorough proposal! I'd like to add a review framed around the *properties* I think any party-naming standard for Canton should satisfy, independent of which design or implementation we ultimately choose. 

My aim is to surface the priorities explicitly so we can debate the requirements first, since I think a lot of the thread is really about which properties to optimize for.



# Assumptions / priorities I'm reviewing from

1. **Security and spoof-resistance is the top priority.** Canton's center of gravity is institutional finance, where preventing impersonation/spoofing attacks matters more than feature breadth.
2. **Resolution up-time/availability should match the ledger it serves.** Name resolution should be roughly as available (or equally) as Canton mainnet itself. There's limited value in a naming layer with weaker uptime than the ledger.
3. **Bias toward proven, well-understood mechanisms, and risk-averse** Where a design choice is uncertain, I'd prefer approaches with a real-world track record over novel mechanisms we'd be validating for the first time on a high-stakes network.
4. **There is no perfect model.** Every option has tradeoffs; the goal is to lean toward the properties we care most about - security, safety, and uptime, and accept costs elsewhere.

These are assumptions, not conclusions. If the WG weights the priorities differently, my assessment changes accordingly.

# Assessment against those priorities

### **Availability** 
I'd suggest up-time belongs among the primary requirements. 

My starting assumption is that a good design should target the availability of the ledger itself. Name resolution that is less available than mainnet has limited value. It would help to understand what availability resolution targets relative to mainnet, and the behavior when a resolver is unreachable.
I recognize that off-ledger resolution was a deliberate choice to limit ACS/Scan load, and that's a real tradeoff worth naming. That said, weighing it against the priorities above, I'd lean toward treating availability (and spoof-resistance) as the harder constraints and cost/footprint as the more negotiable one.

### **Security** 
Given spoof-resistance is a high priority, I'd treat party-name resolution more like:
>  **bank name -> SWIFT/BIC translation** 

than:
>  **domain name -> website IP resolution**. 

SWIFT/BIC works because there is a **single source of truth** for the mapping, ambiguity at the point of use is unacceptable, and you need one authoritative answer before funds move. DNS-style resolution, by contrast, tolerates multiple answers, caching, and failover because the failure modes are different. For Canton's institutional use cases, I don't think we have the same design tolerance for conflicts to exist and be resolved later.
Can we require **conflict prevention**, a single canonical answer per name, rather than resolving conflicts at the point of use? For settlement and counterparty identification, a naming layer probably needs to work like an authoritative registry, not a federated lookup with competing results.

### **Risk posture** 
On a network where security and safety matter most, I'd default **risk-averse**: proven mechanisms over novel ones, especially for what a name resolves to.

cprp-resolution.md
```diff
@@ -0,0 +1,312 @@
...
+
+Status: Draft
+
+Author: Paolo Domenighetti (Freename), [Co-authors TBD]
+
+Created: 2026-08-23
+
+## Abstract
+
+This CIP defines the resolution layer for the Canton Network: how applications resolve human-readable names to Canton Party IDs and how they retrieve associated metadata. It specifies the Fully Qualified Party Name (FQPN) addressing format, the generic Resolver Interface that any identity provider implements, the resolution algorithm, the display conventions for rendered names, the built-in `party-id` resolver, guidelines for writing CIPs for additional resolvers, and a minimum fee model for on-ledger name state.
```
meiersi-da:
```suggestion
This CIP defines the party name resolution layer for the Canton Network: how applications resolve human-readable names to Canton Party IDs and how they retrieve associated metadata. It specifies the Fully Qualified Party Name (FQPN) addressing format, the generic Resolver Interface that any identity provider implements, the resolution algorithm, the display conventions for rendered names, the built-in `party-id` resolver, guidelines for writing CIPs for additional resolvers, and a minimum fee model for on-ledger name state.
```

meiersi-da: Question (may be answered later): why do we need a fee model?

cprp-resolution.md
```diff
@@ -0,0 +1,312 @@
...
+
+This CIP aligns with the CNS 2.0 architecture proposed by Simon Meier (Digital Asset) and the Working Group direction converged over the May–July 2026 meetings. Advanced name verification (a formal T1–T4 trust framework, per-source verification policies) is deferred to future work. Per-source import mechanics (DNS, LEI/vLEI, ENS, cross-chain) are specified in a companion CIP.
```
meiersi-da: This feels out of place in the abstract. I'd push this into "Rationale".

cprp-resolution.md
```diff
@@ -0,0 +1,312 @@
...
+
+## Motivation
+
+Canton participants are identified by cryptographic Party IDs — opaque strings of the form `<prefix>::<68-character-namespace-hash>`. These strings are unusable in any human workflow. CNS 1.0 names (`goldmansachs.unverified.cns`) are self-registered without verification, so they convey no signal about which counterparty they identify.
```
meiersi-da:
```suggestion
Canton participants are identified by cryptographic Party IDs — opaque strings of the form `<prefix>::<68-character-public-key-hash>`. These strings are unusable in any human workflow other than copy-paste. CNS 1.0 names (`goldmansachs.unverified.cns`) are self-registered without verification, so they convey no signal about which counterparty they identify.
```

cprp-resolution.md
```diff
@@ -0,0 +1,312 @@
...
+
+The Identity and Metadata Working Group has identified three concrete gaps:
```
meiersi-da: I'd remove this reference. It reads like a "proof by appeal to authority". I expect a CIP to justify its existence on its own.

cprp-resolution.md
```diff
@@ -0,0 +1,312 @@
...
+
+- P1: Trustworthy human-readable names — applications need to display and accept names that map to specific Canton parties.
+- P2: Off-ledger API endpoint discovery — applications need to find administrative endpoints (for example, the token-admin API for a CIP-56 instrument) keyed on a party identifier. A separate CIP-56 amendment (forthcoming from Digital Asset) will specify endpoint discovery on top of the CN Credentials Standard.
+- P3: Self-published profile information — parties need to publish profile data that displays uniformly across applications, as specified in the Party Profile Credentials CIP (PixelPlex).
+
+This CIP addresses the resolution machinery underlying all three: a uniform addressing model, a pluggable resolver interface, and a shared display convention.
```
meiersi-da: I don't see how P2 requires any part of the CPRP CIP. I also don't think P3 requires it. 

In my mind the motivation for the CPRP CIP is that there are many sources of names, and standardizing the mechanisms for resolving and rendering them avoids committing to a specific one and instead allows the ecosystem to adapt the sources that work best for the usecases at hand whenever the time is right.

cprp-resolution.md
```diff
@@ -0,0 +1,312 @@
...
+
+## Specification
+
+### 1. Terminology
+
+- Party ID. A Canton ledger identity of the form `<prefix>::<namespace>` where the namespace is a 68-character hexadecimal hash of the party administrator's public key. Party IDs may contain additional `::` separators introduced by administrator delegation.
```
meiersi-da: Note that these are 68 characters is an implementation detail that may change. Can you loosen the qualification here?

cprp-resolution.md
```diff
@@ -0,0 +1,312 @@
...
+- Resolver. A software component that provides a mapping between names from an external or internal naming system and Canton Party IDs. Examples: `dns`, `lei`, `ens`, `cns`, `party-id`.
```
meiersi-da:
```suggestion
- Resolver Name. The name of a system that provides a mapping between names from an external or internal naming system and Canton Party IDs. Examples: `dns`, `lei`, `ens`, `cns`, `party-id`.
```

cprp-resolution.md
```diff
@@ -0,0 +1,312 @@
...
+- Name Registry. On-ledger state, replicated in Canton via the CN Credentials Standard, that binds names within a resolver's scope to Party IDs.
```
meiersi-da: Why does it have to be on-ledger for this CIP?

cprp-resolution.md
```diff
@@ -0,0 +1,312 @@
...
+- Registrar. A party approved to operate a name registry (or a subset of it) for a given resolver. Registrars are governance concepts, not part of the FQPN addressing format.
+- FQPN. A Fully Qualified Party Name; see Section 2.
+- Rendered name. The human-readable form of an FQPN shown in application UIs; see Section 6.
```
meiersi-da:
```suggestion
- Rendered name. The human-readable form of an FQPN shown in application UIs.
```

You don't have these sections for the other parts.

cprp-resolution.md
```diff
@@ -0,0 +1,312 @@
...
+
+## Specification
+
+### 1. Terminology
+
+- Party ID. A Canton ledger identity of the form `<prefix>::<namespace>` where the namespace is a 68-character hexadecimal hash of the party administrator's public key. Party IDs may contain additional `::` separators introduced by administrator delegation.
+- Resolver. A software component that provides a mapping between names from an external or internal naming system and Canton Party IDs. Examples: `dns`, `lei`, `ens`, `cns`, `party-id`.
+- Name Registry. On-ledger state, replicated in Canton via the CN Credentials Standard, that binds names within a resolver's scope to Party IDs.
+- Registrar. A party approved to operate a name registry (or a subset of it) for a given resolver. Registrars are governance concepts, not part of the FQPN addressing format.
+- FQPN. A Fully Qualified Party Name; see Section 2.
```
meiersi-da, line 27:
```suggestion
- **Fully Qualified Party Name (FQPN)**: a party name of the form `<network>:<resolver>:<name>`.
```

cprp-resolution.md
```diff
@@ -0,0 +1,312 @@
...
+
+Examples:
+
+- `mainnet:dns:blackrock.com`
+- `mainnet:dns:tmmf.lloyds.com`
+- `mainnet:cns:blackrock.canton`
+- `mainnet:lei:784F5XWPLTWKTBV3E584`
+- `mainnet:party-id:partyHint::1220abcd0000abcd0000abcd0000abcd0000abcd0000abcd0000abcd0000abcd0000`
+
+#### 2.2 Why free-form `<name>`
```
meiersi-da: Move this into the "Rationale" section to keep the spec concise and unambiguous.

See https://github.com/canton-foundation/cips/blob/main/cip-0000/cip-0000.md#cip-format

cprp-resolution.md
```diff
@@ -0,0 +1,312 @@
...
+
+The `<name>` component is a sequence of printable, non-whitespace ASCII characters. It is opaque to the resolution layer and interpreted only by the resolver referenced by the `<resolver>` component. This design allows any external naming syntax to be embedded without escaping: DNS names (`blackrock.com`), CNS names (`blackrock.canton`), LEI codes (`784F5XWPLTWKTBV3E584`), and Canton Party IDs with their `::` separators all fit unchanged in the `<name>` slot.
+
+The three-part format follows the direction proposed by Simon Meier in the June 25, 2026 CNS 2.0 next-steps presentation, after the June 18 Working Group discussion of whether a `registrar` component should appear in the FQPN. The direction: a registrar is a governance concept associated with a resolver, not a first-class addressing component. Adding a registrar slot would multiply icons in the UI, complicate the registrar lifecycle (what happens when a registrar changes for a namespace?), and offer little practical benefit to name discovery.
```
meiersi-da: avoid appealing to authority. Instead provide the reason for why the decision was made this way.

cprp-resolution.md
```diff
@@ -0,0 +1,312 @@
...
+
+Example: `mainnet:party-id:partyHint::1220abcd...`
+
+### 3. The Resolver Interface
+
+Every resolver — built-in or third-party — implements the following logical interface. HTTP and gRPC mappings are conventional; see Section 8.
+
+| Method | Inputs | Output | Purpose |
+|--------|--------|--------|---------|
+| `resolve` | name | ResolutionResult | Forward lookup: name → Party ID + metadata |
```
meiersi-da: what about fuzzy resolution for auto-complete?

cprp-resolution.md
```diff
@@ -0,0 +1,312 @@
...
+
+The off-ledger placement is a deliberate design choice to limit ACS growth on the DSO party and to keep the Scan API surface narrow. The tradeoff is that the resolution layer's availability depends on resolver operator infrastructure rather than being an inherent property of the ledger. Applications with settlement-critical dependencies on name resolution SHOULD provision or contract for resolver infrastructure at the same availability standard as their ledger connectivity.
+
+### 9. On-Ledger Representation
+
+Name registrations are encoded as standard CN Credentials, using the credential registry defined by the CN Credentials Standard CIP (Digital Asset). This CIP does not introduce custom Daml templates.
```
meiersi-da: I would explicitly not require this.

meiersi-da (commented):
I have not yet reviewed CPRP in completeness, and have not done anything for the others. Sending you these comments as I have to go into another meeting.

cprp-resolution.md
```diff
@@ -0,0 +1,312 @@
...
+- Read traffic (resolution queries) is free.
+- Write traffic (registrations, updates, renewals) is paid via a fee.
+
+The recommended implementation reuses the extended-duration credential mechanism introduced by the CN Credentials Standard CIP (proposed default fee of $1/year, first 90 days free, paid by CC burn requiring authorization from the funds owner and the issuer or holder). Registrars MAY charge above the minimum for higher-value names; this pricing latitude is one mechanism for reducing name squatting. Full tokenomics is out of scope for this CIP and expected to be addressed in a future governance CIP.
+
+### 11. Guidelines for Writing CIPs for Additional Resolvers
+
+Any project may propose a new resolver via its own CIP. Such a CIP MUST specify:
+
+- The resolver's `<resolver>` identifier as it appears in FQPNs.
```
meiersi-da:
```suggestion
- The unique `<resolver name>` of the resolver as it appears in FQPNs.
```

cprp-resolution.md
```diff
@@ -0,0 +1,312 @@
...
+- The `<name>` syntax accepted by this resolver.
+- The naming authority and governance under which registrars operate.
+- The verification procedure by which the resolver establishes name-to-Party-ID bindings.
+- The credential encoding used to publish on-ledger name registrations.
+- The re-verification cadence, if the underlying authority state can change externally.
```
meiersi-da: why do we need these?

cprp-resolution.md
```diff
@@ -0,0 +1,312 @@
...
+
+### 10. Fees
+
+Name registration on-ledger consumes state and warrants a minimum fee model to prevent unbounded growth and to fund registrar operations. This CIP specifies the shape of the fee model, following the direction confirmed by Simon Meier at the WG meeting of July 9, 2026:
+
+- Fees are paid per registered name.
+- Read traffic (resolution queries) is free.
+- Write traffic (registrations, updates, renewals) is paid via a fee.
+
+The recommended implementation reuses the extended-duration credential mechanism introduced by the CN Credentials Standard CIP (proposed default fee of $1/year, first 90 days free, paid by CC burn requiring authorization from the funds owner and the issuer or holder). Registrars MAY charge above the minimum for higher-value names; this pricing latitude is one mechanism for reducing name squatting. Full tokenomics is out of scope for this CIP and expected to be addressed in a future governance CIP.
```
meiersi-da: Why do we need to standardize on these? The one thing that we should probably standardize on is to set the expectation that name resolution should by default be free within a fair use policy.

cprp-resolution.md
meiersi-da: It limits the implementation freedom without clear gains.

What we do want to standardize on is how the names are represented (FQPNs), resolved, and rendered; and how additional name sources can be integrated into Canton Network by defining new resolvers or extending existing ones.

cprp-resolution.md
```diff
@@ -0,0 +1,312 @@
...
+| Profile summary | Hover tooltip, popover | Rendered name + resolver + published profile claims |
+| Full profile | Explorer profile page | Complete credential set, resolver history, changelog |
+
+The full profile MAY be hosted by any explorer, not exclusively the Canton Foundation's Scan.
+
+### 7. Trust and Verification
+
+This CIP intentionally does not define a formal trust framework. The initial trust model is minimal:
+
+- The DSO party issues credentials naming approved TLD registry admins and approved registrars per TLD (see the CNS 2.0 CIP led by Digital Asset).
```
meiersi-da: TLD's are a CNS concept. They are not part of FQPNs AFAIS. So we shouldn't standardize these here.

cprp-resolution.md
```diff
@@ -0,0 +1,312 @@
...
+
+Assets do not receive their own FQPNs; they are addressed by admin + id per the token standard, and rendered via this convention.
+
+#### 6.2 Three Rendering Contexts
+
+| Context | Where used | Content |
+|---------|------------|---------|
+| Inline | Transaction lists, counterparty fields | `<icon> <display-name>` |
+| Profile summary | Hover tooltip, popover | Rendered name + resolver + published profile claims |
+| Full profile | Explorer profile page | Complete credential set, resolver history, changelog |
```
meiersi-da: These two belong the the party profile CIP AFAIS.

cprp-resolution.md
```diff
@@ -0,0 +1,312 @@
...
+
+The full profile MAY be hosted by any explorer, not exclusively the Canton Foundation's Scan.
```
meiersi-da: The profile will likely not be hosted on the SV node's Scan instance. Let's remove that hint.

cprp-resolution.md
```diff
@@ -0,0 +1,312 @@
...
+Rendered names follow a shared convention: a source icon followed by a human-readable ASCII string.
+
+```
+<icon> <display-name>
+```
+
+The icon identifies the source of the name and is computed from the FQPN:
+
+- For external-system resolvers (`dns`, `lei`, `ens`) and the built-in `party-id` resolver, there is one standardized icon per resolver.
+- For the `cns` resolver, the icon is the logo of the TLD under which the name is registered. Registrars self-publish their desired logo for their TLD, and the logo is registered via a DSO-issued credential (`cip-cns/tld-logo-url`, per the CNS 2.0 CIP). At render time, the application retrieves the logo URL from the DSO Credentials Registry and displays `<img src=logoUrl> <display-name>`.
```
meiersi-da: Let's not define the CNS resolver in this standard.

cprp-resolution.md
```diff
@@ -0,0 +1,312 @@
...
+| Code | Meaning |
+|------|---------|
+| 1000 | Name not found |
+| 1001 | Resolver temporarily unavailable |
+| 1002 | Malformed FQPN |
+| 1003 | Network mismatch |
+
+### 4. Name Resolution Algorithm
+
+An application resolving a user-typed name follows a small, deterministic algorithm. The default is intentionally simple; multi-source composition is an optional advanced feature (Section 5).
```
meiersi-da: Can we avoid the optional feature? Optional features lead to ecosystem fragmentation into the parts that do support it and the ones that don't. Let's focus on the minimal parts that are required to work with FQPNs in their ASCII and graphical rendering.

cprp-resolution.md
```diff
@@ -0,0 +1,312 @@
...
+#### 8.1 API
+
+The Resolver Interface (Section 3) is exposed by convention as an HTTPS + gRPC service.
+
+| Endpoint | Method | Purpose |
+|----------|--------|---------|
+| `/v1/resolve` | POST | Single forward lookup |
+| `/v1/resolve/batch` | POST | Batched forward lookup |
+| `/v1/resolve/reverse` | POST | Reverse lookup |
+| `/v1/changelog` | GET (SSE) | Subscribe to changes |
```
meiersi-da: A reference to an OpenAPI file for the full API would be great to have. It removes the ambiguity and creates value by precisely defining the API that we'd expect to mediate between apps and name resolvers.

meiersi-da: You could create one as a PR against splice.

meiersi-da (commented):
Thanks a lot for all the polishing work. While I think it has all the right pieces it still contains too many of them AFAIS. Less is more. I believe this comment sums up well what I was looking to standardize as part of this CIP: https://github.com/canton-foundation/cips/pull/171/changes#r3854762146

cprp-resolution.md
```diff
@@ -0,0 +1,147 @@
...
+
+Status: Draft
+
+Author: Paolo Domenighetti (Freename), [Co-authors TBD]
+
+Created: 2026-08-27
+
+## Abstract
+
+This CIP standardizes four things and deliberately nothing more: how human-readable party names are represented on the Canton Network (the Fully Qualified Party Name format), how they are resolved to Canton Party IDs, how they are rendered in application UIs, and how additional name sources are integrated by defining new resolvers or extending existing ones.
```
meiersi-da:
```suggestion
This CIP defines a uniform approach for naming parties on the Canton network.
The approach allows integrating the many naming systems used by the organizations on the Canton Network regardless of whether they are native to the Canton Network (like CNS) or externally managed (like DNS or LEI).
The approach is based on standardizing the following four concepts as part of this CIP: how human-readable party names are represented on the Canton Network (the Fully Qualified Party Name format), how they are resolved to Canton Party IDs, how they are rendered in application UIs, and how additional naming systems are integrated by defining new resolvers or extending existing ones.
```

Note: let's decide on "naming system" over "name source" and use consistently. I'm leaning towards the former as it seems more general.

cprp-resolution.md
```diff
@@ -0,0 +1,147 @@
...
+
+It aligns with the CNS 2.0 architecture proposed by Simon Meier (Digital Asset) and the Working Group direction converged over the May–August 2026 meetings.
```
meiersi-da: Push to rationale. Make the CIP earn its merit based on its text.

cprp-resolution.md
```diff
@@ -0,0 +1,147 @@
...
+
+## Motivation
+
+Canton participants are identified by cryptographic Party IDs of the form `<prefix>::<68-character-namespace-hash>` — unusable in human workflows. CNS 1.0 names (`name.unverified.cns`) are self-registered without verification and convey no signal about which counterparty they identify. Applications need to display trustworthy human-readable names, accept them as input, and discover metadata keyed on a party — uniformly across the network. This CIP provides the shared representation, resolution behavior, and rendering conventions that make that possible, while leaving every implementation choice to implementers.
```
meiersi-da:
```suggestion
Canton Network participants are identified by cryptographic Party IDs of the form `<prefix>::<68-character-namespace-hash>`, which are unusable in human workflows other than simple copy-and-pasting. CNS 1.0 names (`name.unverified.cns`) are self-registered without verification and convey no signal about which counterparty they identify. Applications need to display memorable human-readable names and accept them as input uniformly across the network. This CIP provides the shared representation, resolution behavior, and rendering conventions that make that possible, while leaving ample implementation freedom to implementers.
```

cprp-resolution.md
```diff
@@ -0,0 +1,147 @@
...
+
+### 5. Integrating Additional Name Sources
+
+New name sources enter the Canton Network either by defining a new resolver or by extending an existing one, in each case via a dedicated CIP. Such a CIP MUST specify:
+
+- The resolver identifier as it appears in FQPNs (or the existing resolver being extended).
+- The `<name>` syntax the resolver accepts.
+- The naming authority and the governance under which its registrars operate.
+- The procedure by which name-to-Party-ID bindings are established and, where the underlying authority can change externally, the re-verification cadence.
+- The on-ledger credential encoding of name registrations, using the CN Credentials Standard. Names MUST be materialized as on-ledger credentials so that Daml code can reference them (import model, per the July 9, 2026 WG direction); reference-only integration without on-ledger state is acceptable only where no Daml logic depends on the name.
```
meiersi-da: I would make this optional. Names have ample value for display purposes only.

meiersi-da: I'm even inclined to completely remove this CIP. It seems to be more of an aspect of machine-checked identity verification than an aspect of making it easy for humans to interact with party-ids.

cprp-resolution.md
```diff
@@ -0,0 +1,147 @@
...
+- The rendered-name convention, if it differs from `<name>` verbatim.
+
+Registration credentials SHOULD carry at minimum the canonical FQPN, the network, and a validity horizon as claims. The claim-key prefix used by this document family (`cprp/`) is a working prefix, to be renamed `cip-<nr>/` on number assignment, consistent with `cip-cns/*`.
+
+The first such source CIP is the Canton Imported Names CIP (Decentralized DNS Name Importer), which specifies DNS import in full and serves as the blueprint for LEI/vLEI, ENS, email, and further importers.
```
meiersi-da: Don't precommit to another CIP in this CIP ;-)

meiersi-da: let's remove this

cprp-resolution.md
```diff
@@ -0,0 +1,147 @@
...
+- `mainnet:lei:784F5XWPLTWKTBV3E584`
+- `mainnet:party-id:partyHint::1220abcd0000abcd0000abcd0000abcd0000abcd0000abcd0000abcd0000abcd0000`
+
+The `<name>` component is opaque to this standard and interpreted only by the resolver. This lets any external naming syntax — DNS dots, LEI codes, Party IDs with their `::` separators — embed without escaping.
+
+#### 2.2 Network Discrimination
+
+The `network` component is mandatory. TestNet names MUST NOT be confusable with MainNet names. Applications supporting multiple networks SHOULD display the network.
+
+#### 2.3 The `party-id` Resolver (Built-in)
```
meiersi-da: this is a resolution aspect not a representation aspect

cprp-resolution.md
```diff
@@ -0,0 +1,147 @@
...
+The `party-id` resolver is defined by this CIP and available on every network. Its `<name>` is the Canton Party ID verbatim. It guarantees that every Canton party always has at least one FQPN, before any name registration, with no external state and no trust beyond the Party ID itself.
+
+### 3. Resolution
+
+#### 3.1 Resolver Operations
+
+Every resolver provides two logical operations. How they are transported and exposed is implementation-defined.
+
+- resolve: name → the Party ID the name maps to, plus associated metadata claims and a validity horizon.
+- reverseResolve: Party ID → the names this resolver knows for it.
```
meiersi-da: Here I'm inclined to give an actual HTTP OpenAPI spec. Otherwise name providers and consumers cannot act independently and have to do more coordination. Removing this coordination need is the goal of these standardization CIPs.

cprp-resolution.md
```diff
@@ -0,0 +1,147 @@
...
+
+#### 3.2 Forward Resolution
+
+An application resolving a user-typed name matches it against each resolver in its configured, ordered list and returns the first successful result. Per-resolver ignore rules (patterns on the `<name>` component) MAY skip a resolver for matching names, addressing overlaps between resolvers when the application wants a specific delegation.
+
+#### 3.3 Preferred-Name Resolution (Reverse Direction)
+
+Given a Party ID, an application computes the preferred FQPN to display:
+
+1. Retrieve the party's preferred-name credential from the Credentials Registry: issuer = holder = the party, claim `cip-cns/preferred-name` (claim key per the CNS 2.0 CIP).
```
meiersi-da: Interesting. I wasn't aware of this dependency. Makes sense to have. Note though that it's not the CNS CIP, but this CIP that defines how parties publish their preferred names. 

That said: I'm kind of inclined to actually push this into the party profile CIP. It seems like another kind of self-published party metadata. wdyt?

@waynecollier-da wdyt?

cprp-resolution.md
```diff
@@ -0,0 +1,147 @@
...
+2. Interpret the claim value as an FQPN and resolve it forward per Section 3.2, verifying it resolves back to the same Party ID. This round-trip verification prevents a party from claiming a name it does not hold.
+3. If verified, display that name.
+4. Otherwise, perform reverse resolution across the configured resolvers and select one result by the application's preference order — by default, lexicographic sort by (application priority override, user priority, application priority default).
+
+### 4. Rendering
```
meiersi-da: I'd suggest to move this right after the representation. I think it would also be good to have two subsections: one for the ASCII Rendering, which just states that it is verbatim, and one for the GUI Rendering.

cprp-resolution.md
```diff
@@ -0,0 +1,147 @@
...
+
+Rendered names follow a shared convention: a source icon followed by the human-readable name.
+
+```
+<icon> <display-name>
+```
+
+- For `dns`, `lei`, `ens`, and `party-id`, there is one standardized icon per resolver.
+- For `cns`, the icon is the logo of the TLD under which the name is registered. Registrars self-publish their TLD logo, registered via a DSO-issued credential (`cip-cns/tld-logo-url`, per the CNS 2.0 CIP); applications retrieve the logo URL from the Credentials Registry at render time.
```
meiersi-da: Avoid the forward reference. Instead structure the icon resolution algorithm such that resolvers can specify both how the icon and the name is determined.

meiersi-da: Also: I'd suggest to add a screenshot of the Etherscan and its <icon> <name> format in the rationale to justify this decision giving prior art.

meiersi-da (commented):
Comments from a review of the text up to Backwards Compatibility.

It starts to be very nicely focused. I understand it's surprisingly small, but that's IMHO a feature.

cprp-resolution.md
meiersi-da: Note: @pdomenighetti and myself agree that this should be pushed to party profiles.

cprp-resolution.md
```diff
@@ -0,0 +1,220 @@
+Number: CIP-XXXX
+
+Title: Canton Party Name Resolution Standard
+
+Status: Draft
+
+Author: Paolo Domenighetti (Freename), [Co-authors TBD]
+
+Created: 2026-08-29
```
meiersi-da, line 9: consider adjusting to the default preamble defined here: https://github.com/canton-foundation/cips/blob/main/cip-XXXX/cip-XXXX.md?plain=1

cprp-resolution.md
```diff
@@ -0,0 +1,220 @@
...
+
+## Motivation
+
+Canton Network participants are identified by cryptographic Party IDs of the form `<prefix>::<68-character-namespace-hash>`, which are unusable in human workflows other than simple copy-and-pasting. CNS 1.0 names (`name.unverified.cns`) are self-registered without verification and convey no signal about which counterparty they identify. Applications need to display memorable human-readable names and accept them as input uniformly across the network. This CIP provides the shared representation, resolution behavior, and rendering conventions that make that possible, while leaving ample implementation freedom to implementers.
```
meiersi-da, line 17:
```suggestion
Canton Network participants are identified by cryptographic Party IDs of the form `<prefix>::<68-character-namespace-hash>`, which are unusable in human workflows other than simple copy-and-pasting. Applications need to display memorable human-readable names and accept them as input uniformly across the network. This CIP provides the shared representation, resolution behavior, and rendering conventions that make that possible, while leaving ample implementation freedom for implementing or integrating naming systems.
```

cprp-resolution.md
```diff
@@ -0,0 +1,220 @@
...
+
+## Specification
+
+### 1. Terminology
+
+- Party ID. A Canton ledger identity of the form `<prefix>::<namespace>`, where the namespace is a 68-character hexadecimal hash of the party administrator's public key. Party IDs may contain additional `::` separators.
```
meiersi-da, line 23:
```suggestion
- **Party ID**: a Canton ledger identity of the form `<prefix>::<namespace>`, where the namespace is the hash of the party owners public key.
```

let's just define the terms here. if we need the deleted background knowledge, then let's provide that in a **Rationale > Technical Background** section

meiersi-da: nit: I personally find the

> - **bold term**: definition

style easier to parse, and have been using it in all my CIPs. Consider switching to it.

cprp-resolution.md
```diff
@@ -0,0 +1,220 @@
...
+- Naming system. A system that manages human-readable names, native to the Canton Network (e.g. CNS) or externally managed (e.g. DNS, LEI, ENS).
+- Resolver. A component that maps names from a naming system to Canton Party IDs. Examples: `dns`, `lei`, `ens`, `cns`, `party-id`.
+- Registrar. A party approved to allocate names within a naming system. Registrars are a governance concept defined by each naming system's CIP; they are not part of the addressing format.
```
meiersi-da, line 26:
```suggestion
- Registrar. An organization approved to allocate names within a naming system. Registrars are a governance concept defined by each naming system's CIP; they are not part of the addressing format.
```

there are registrars in systems that are not Canton native

cprp-resolution.md
```diff
@@ -0,0 +1,220 @@
...
+- FQPN. A Fully Qualified Party Name; see Section 2.
+
+### 2. Representation — the Fully Qualified Party Name
+
```
meiersi-da, line 30:
```suggestion

Fully qualified party names provide a uniform structure for naming parties in Canton.

```

cprp-resolution.md
```diff
@@ -0,0 +1,220 @@
...
+#### 3.2 GUI Rendering
+
+In graphical UIs, rendered names follow a shared convention: a source icon followed by the human-readable display name.
+
+```
+<icon> <display-name>
+```
+
+How the icon and the display name are determined is specified by each naming system's CIP (see Section 5). This CIP fixes only the convention: one icon identifying the source of the name, followed by the display name in the naming system's canonical form (e.g. `blackrock.com`, `blackrock.canton`, a truncated Party ID prefix).
```
meiersi-da, line 71:
```suggestion
How the icon and the display name are determined is specified by each resolver's own CIP (see Section 5). Here we only fix the convention: one icon identifying the source of the name, followed by the display name in the naming system's canonical form (e.g. `blackrock.com`, `blackrock.canton`, a truncated Party ID).
```

cprp-resolution.md
```diff
@@ -0,0 +1,220 @@
...
+
+Fallback chain when no name resolves for a party: CNS 1.0 entry → Party ID prefix → truncated Party ID.
+
+Assets are not named by FQPNs. An instrument identified by the token standard's `InstrumentId {admin: Party, id: Text}` renders as `<symbol> by <admin-rendered-name>` (e.g. `CC by <cns icon> dso.cns`), where the symbol is published by the admin through its token-standard metadata endpoint (`/registry/metadata/v1/instruments/{id}`), discovered via the admin's credentials.
```
meiersi-da, line 75: This feels a bit off here. It makes sense to define, but should probably move into its own section 3.3

cprp-resolution.md
```diff
@@ -0,0 +1,220 @@
...
+
+In graphical UIs, rendered names follow a shared convention: a source icon followed by the human-readable display name.
+
+```
+<icon> <display-name>
+```
+
+How the icon and the display name are determined is specified by each naming system's CIP (see Section 5). This CIP fixes only the convention: one icon identifying the source of the name, followed by the display name in the naming system's canonical form (e.g. `blackrock.com`, `blackrock.canton`, a truncated Party ID prefix).
+
+Fallback chain when no name resolves for a party: CNS 1.0 entry → Party ID prefix → truncated Party ID.
```
meiersi-da, line 73: This seems like a left-over stub. It does though hint at missing sections in resolution that answer these questions

- how should application make use of the resolvers?
  - configure the resolvers they want to use and in what order
  - they can always use the fallback `party-id` resolver
- how should they deal with duplicate names?
  - by default use the one from the resolver they prefer
  - if there are multiple names, then consider displaying them to the user where helpful (e.g. on hover)

meiersi-da: Obviously better titles should be chosen than just the question ;-)

cprp-resolution.md
```diff
@@ -0,0 +1,220 @@
...
+
+Assets are not named by FQPNs. An instrument identified by the token standard's `InstrumentId {admin: Party, id: Text}` renders as `<symbol> by <admin-rendered-name>` (e.g. `CC by <cns icon> dso.cns`), where the symbol is published by the admin through its token-standard metadata endpoint (`/registry/metadata/v1/instruments/{id}`), discovered via the admin's credentials.
+
+### 4. Resolution
+
+#### 4.1 Resolver Operations
+
+Every resolver exposes two operations over HTTPS, specified by the following OpenAPI definition so that name providers and consumers can act independently without pairwise coordination.
+
+```yaml
```
meiersi-da, line 83: Embedding the OpenAPI spec in the CIP makes it really hard to evolve. What we usually do is to refer to a DRAFT PR that adds it in the right place to splice.

`naming-standards/party-name-resolution` seems like a decent location within Splice for this definition

cprp-resolution.md
```diff
@@ -0,0 +1,220 @@
...
+          schema: { type: string, enum: [mainnet, testnet, devnet, localnet] }
+      responses:
+        "200":
+          description: Resolved
+          content:
+            application/json:
+              schema: { $ref: "#/components/schemas/ResolutionResult" }
+        "404":
+          description: Name not found
+  /v0/reverse-resolve:
```
meiersi-da, line 109: Seeing these APIs makes one ask the question as to why we don't implement them in Scan for the existing CNS 1.0 entries. I'd say we don't do that here, as the right way to do so depends on how the CNS resolver will work going forward, which will be defined in its own CIP. That CIP should then also be the one that explains how Scan serves these APIs.

meiersi-da: If you agree, then could you add a corresponding entry in the rationale section?

cprp-resolution.md
```diff
@@ -0,0 +1,220 @@
...
+      summary: List the names this resolver knows for a Party ID
+      parameters:
+        - name: party
+          in: query
+          required: true
+          schema: { type: string }
+        - name: network
+          in: query
+          required: true
+          schema: { type: string, enum: [mainnet, testnet, devnet, localnet] }
```
meiersi-da, line 120: Missing: limits and pagination

cprp-resolution.md
```diff
@@ -0,0 +1,220 @@
...
+          content:
+            application/json:
+              schema: { $ref: "#/components/schemas/ResolutionResult" }
+        "404":
+          description: Name not found
+  /v0/reverse-resolve:
+    get:
+      summary: List the names this resolver knows for a Party ID
+      parameters:
+        - name: party
```
meiersi-da, line 113: Missing for all params: `description`

cprp-resolution.md
```diff
@@ -0,0 +1,220 @@
...
+  schemas:
+    ResolutionResult:
+      type: object
+      required: [fqpn, party_id, display_name]
+      properties:
+        fqpn: { type: string }
+        party_id: { type: string }
+        display_name: { type: string }
+        metadata: { type: object, additionalProperties: true }
+        valid_until: { type: string, format: date-time }
```
meiersi-da, line 139: missing: description explaining the semantics

cprp-resolution.md
```diff
@@ -0,0 +1,220 @@
...
+        "200":
+          description: Known names (possibly empty)
+          content:
+            application/json:
+              schema:
+                type: array
+                items: { $ref: "#/components/schemas/ResolutionResult" }
+components:
+  schemas:
+    ResolutionResult:
```
meiersi-da, line 131:
```suggestion
    ReverseResolutionResult:
```

for this one it doesn't make sense to repeat the party-id for every name does it?

cprp-resolution.md
```diff
@@ -0,0 +1,220 @@
...
+
+Every resolver exposes two operations over HTTPS, specified by the following OpenAPI definition so that name providers and consumers can act independently without pairwise coordination.
+
+```yaml
+openapi: 3.0.3
+info:
+  title: Canton Party Name Resolution API
+  version: 0.1.0
+paths:
+  /v0/resolve:
```
meiersi-da, line 89: wdyt about adding an endpoint to support auto-complete? It's up to the server how to do that, but it does require a different result type

meiersi-da: and a limit param

cprp-resolution.md
```diff
@@ -0,0 +1,220 @@
...
+components:
+  schemas:
+    ResolutionResult:
+      type: object
+      required: [fqpn, party_id, display_name]
+      properties:
+        fqpn: { type: string }
+        party_id: { type: string }
+        display_name: { type: string }
+        metadata: { type: object, additionalProperties: true }
```
meiersi-da, line 138: explain in description how this is intended to be used

cprp-resolution.md
meiersi-da: better inline here so people using codegens still have access to this info

cprp-resolution.md
```diff
@@ -0,0 +1,220 @@
...
+        valid_until: { type: string, format: date-time }
+```
+
+Deployment, scaling, caching, and authentication are implementation-defined. A `ResolutionResult` whose `valid_until` has passed MUST be re-resolved before use.
+
+#### 4.2 Forward Resolution
+
+An application resolving a user-typed name matches it against each resolver in its configured, ordered list and returns the first successful result. Per-resolver ignore rules (patterns on the `<name>` component) MAY skip a resolver for matching names, addressing overlaps between resolvers when the application wants a specific delegation.
```
meiersi-da, line 146: So this is about going from just `name` to an FQPN, correct? It seems sort of helpful, but it also misses the aspect that users typically type a fragment (or prefix) of a name.

cprp-resolution.md
```diff
@@ -0,0 +1,220 @@
...
+
+#### 4.3 The `party-id` Resolver (Built-in)
```
meiersi-da, line 148: I'd suggest to move this into its own section "The `party-id` resolver"

meiersi-da: Also explain that there is no server of the OpenAPI spec for this one.

meiersi-da: This actually makes me question my suggestion to define an OpenAPI spec. It seems that what we really require is that it is easy to compose the results from different resolvers client-side. We might though be able to do that using a simple spec that just defines input and output parameters of the three methods: resolveName, fuzzySearch, resolvePartyId

wdyt?

cprp-resolution.md
```diff
@@ -0,0 +1,220 @@
...
+
+The `party-id` resolver is defined by this CIP and available on every network. Its `<name>` is the Canton Party ID verbatim; it resolves trivially and reverse-resolves any Party ID to itself. It guarantees that every Canton party always has at least one FQPN, before any name registration, with no external state and no trust beyond the Party ID itself.
+
+#### 4.4 Preferred-Name Resolution (Reverse Direction)
+
+Given a Party ID, an application computes the preferred FQPN to display:
+
+1. Retrieve the party's preferred-name credential from the Credentials Registry: issuer = holder = the party, claim `cprp/preferred-name` (defined by this CIP).
```
meiersi-da, line 156:
```suggestion
1. Retrieve the party's preferred-name credential from the DSO Credentials Registry: issuer = holder = the party, claim `cprp/preferred-name` (defined by this CIP).
```

cprp-resolution.md
```diff
@@ -0,0 +1,220 @@
...
+
+- The resolver identifier as it appears in FQPNs (or the existing resolver being extended). Resolver identifiers MUST be unique across CIPs; the CIP process itself acts as the registry of resolver identifiers.
+- The `<name>` syntax the resolver accepts.
+- The naming authority and the governance under which its registrars operate.
+- The procedure by which name-to-Party-ID bindings are established and, where the underlying authority can change externally, the re-verification cadence.
+- How the source icon and the display name are determined for GUI rendering (Section 3.2).
+- The on-ledger credential encoding of name registrations, where the naming system provides one, using the CN Credentials Standard. On-ledger materialization is optional: names have ample value for display purposes alone; naming systems whose names must be referenced by Daml code materialize them as on-ledger credentials.
+
+Registration credentials, where used, SHOULD carry at minimum the canonical FQPN, the network, and a validity horizon as claims. The claim-key prefix used by this CIP (`cprp/`) is a working prefix, to be renamed `cip-<nr>/` on number assignment.
```
meiersi-da, line 172: Would you mint pushing this into its own subsection. Right now it is confusing to see optional bullet points in the MUST list.

cprp-resolution.md
```diff
@@ -0,0 +1,220 @@
...
+
+### 6. Backward Compatibility
+
+Additive. CNS 1.0 names are preserved and resolve under the `cns` resolver in their existing display form. Party IDs continue to work as they always have; the `party-id` resolver exposes them as FQPNs. Applications that do not adopt this CIP are unaffected.
```
meiersi-da, line 176: They do not. The `cns` resolver is not define as part of this CIP.

cprp-resolution.md
```diff
@@ -0,0 +1,220 @@
...
+
+## Rationale
+
+### Alignment
+
+This CIP aligns with the CNS 2.0 architecture proposed within the Canton Identity and Metadata Working Group and the direction converged over the May–August 2026 meetings.
```
meiersi-da, line 182: add link to https://docs.google.com/presentation/d/1-9OA7IPnvRPE4w1MwaQ6eAPSSqSUAXiNfdExCzLmkO8/edit?slide=id.g3d69844f64f_1_1675#slide=id.g3d69844f64f_1_1675

cprp-resolution.md
```diff
@@ -0,0 +1,220 @@
...
+
+### Why exactly these four concepts
+
+Representation, resolution, rendering, and integration are what require network-wide agreement: two applications must parse the same FQPN, obtain the same binding, show the user a consistent picture, and let new naming systems plug in predictably. Everything else — deployment, scaling, economics, composition policies — can differ between implementations without harming interoperability, so this CIP leaves it free.
```
meiersi-da, line 186: Note that asset names are not really covered by these four concepts. It might be OK though.

cprp-resolution.md
```diff
@@ -0,0 +1,220 @@
...
+
+Name-selection UIs in the wild are always context-scoped; a single ordered resolver list with ignore rules covers them. Cross-source result composition was part of earlier drafts of this proposal and has been removed: it standardized behavior that applications can layer on privately if they need it, at the cost of a materially larger standard.
+
+### Why round-trip verification for preferred names
+
+A preferred-name claim is self-issued. Verifying that the claimed FQPN actually resolves back to the claiming party costs one forward resolution and eliminates the entire class of display-name spoofing via false preference claims.
+
+### Prior art for icon-plus-name rendering
+
+The `<icon> <display-name>` convention follows established block-explorer practice. Etherscan's transaction lists render known addresses as a source icon followed by a human-readable name in place of the raw hex — exchange tags, contract labels, and ENS names side by side in the same table — which has proven legible and spoof-resistant at scale:
```
meiersi-da, line 210:
```suggestion
The `<icon> <display-name>` convention follows established block-explorer practice. Etherscan's transaction lists render known addresses as a source icon followed by a human-readable name in place of the raw hex — exchange tags, contract labels, and ENS names side by side in the same table. We copy this proven design decision:
```

meiersi-da (commented):
Thanks a lot @pdomenighetti for all the good improvements; and sorry for the large number of comments. Most of them should be easy to apply. The only major concern seems to be how to specify the different resolution functions and how they are composed across resolvers.

PR #191, opened 2026-04-21
cip-105: Clarify mandatory vesting

stas-sbi:


waynecollier-da (approved):
Looks good to me; thanks for adding these clarifications @stas-sbi

PR #192, opened 2026-04-22
Update cip-0111.md

pedrodneves:


hythloda (approved):

cip-0111/cip-0111.md
stas-sbi, line 6: This is a process, not governance
```suggestion
Type: Process
```

PR #204, opened 2026-05-04
DRAFT: Canton Network Credentials Standard

meiersi-da:
New home for the previous GDoc version in [CIP DRAFT - CN credentials standard](https://docs.google.com/document/d/1wsjfsAfEnW_nPJcZ8SsV0PyVsXPgqy2tOoBtyIWulMs/edit?userstoinvite=support%40smalltimedevs.com&sharingaction=manageaccess&role=reader&tab=t.0).

Please add new comments in the PR review here.


cn-credentials-standard/cip-canton-network-credentials-standard.md
```diff
@@ -0,0 +1,525 @@
...
+
+This CIP is licensed under CC0-1.0: [Creative Commons CC0 1.0 Universal](https://creativecommons.org/publicdomain/zero/1.0/)
+
+## Changelog
+
+Jan 9, 2026: wrote first draft
+
+## Appendix
+
+### Open Comment Threads
```
meiersi-da, line 477: TODO: look into how to make on- and off-ledger credentials usable in the same way, as pointed out by Luis Marado.

We may be able to reuse some of the ideas from: https://github.com/kaikodata/canton-data-standard/pull/14/changes#diff-436aced28f23276f5a25d27fd242f38d15807dd2c7edc2b094cd628fd210f779R42

Noosphere-314:
@simonletort-da, I read the draft lifecycle and the linked registry prototype with one kind of credential in mind: an eligibility claim that an issuer makes about a holder, which an app then acts on off-ledger.

The interface lets the holder archive a credential, and updates through the factory are controlled by holder and issuer together. In the registry prototype, the issuer can archive on its own through `AnsCredential_Expire` once `expiresAt` has passed, which defaults to 90 days after creation; the DSO has a separate path for records that are invalid under its rules.

If an issuer needs to withdraw an eligibility claim before expiry and the holder does not take part, is the revocation signal a consumer should rely on meant to be defined by each registry, or is an issuer-initiated withdrawal in scope for the standard?

I realise KYC-specific claims are outside this draft, so I am asking about issuer-issued credentials in general. The linked registry PR is closed, so it may no longer reflect the intended implementation.

meiersi-da:
> @meiersi-da, I read the draft lifecycle and the linked registry prototype with one kind of credential in mind: an eligibility claim that an issuer makes about a holder, which an app then acts on off-ledger.
> 
> The interface lets the holder archive a credential, and updates through the factory are controlled by holder and issuer together. In the registry prototype, the issuer can archive on its own through `AnsCredential_Expire` once `expiresAt` has passed, which defaults to 90 days after creation; the DSO has a separate path for records that are invalid under its rules.
> 
> If an issuer needs to withdraw an eligibility claim before expiry and the holder does not take part, is the revocation signal a consumer should rely on meant to be defined by each registry, or is an issuer-initiated withdrawal in scope for the standard?

I would not expect the registry to define the interpretation, but the apps consuming the credentials and the apps producing the credentials. The registry is just infrastructure. Furthermore, if the authorization model for the DSO Credential Registry is not suitable, then the issuer could also run their own registry.
 
> I realise KYC-specific claims are outside this draft, so I am asking about issuer-issued credentials in general. The linked registry PR is closed, so it may no longer reflect the intended implementation.

The closed PR does reflect what's intended to be implemented to a large extent. The details will be worked out by the contributors that take this CIP to its conclusion.

Noosphere-314:
> > @meiersi-da, I read the draft lifecycle and the linked registry prototype with one kind of credential in mind: an eligibility claim that an issuer makes about a holder, which an app then acts on off-ledger.
> > The interface lets the holder archive a credential, and updates through the factory are controlled by holder and issuer together. In the registry prototype, the issuer can archive on its own through `AnsCredential_Expire` once `expiresAt` has passed, which defaults to 90 days after creation; the DSO has a separate path for records that are invalid under its rules.
> > If an issuer needs to withdraw an eligibility claim before expiry and the holder does not take part, is the revocation signal a consumer should rely on meant to be defined by each registry, or is an issuer-initiated withdrawal in scope for the standard?
> 
> I would not expect the registry to define the interpretation, but the apps consuming the credentials and the apps producing the credentials. The registry is just infrastructure. Furthermore, if the authorization model for the DSO Credential Registry is not suitable, then the issuer could also run their own registry.
> 
> > I realise KYC-specific claims are outside this draft, so I am asking about issuer-issued credentials in general. The linked registry PR is closed, so it may no longer reflect the intended implementation.
> 
> The closed PR does reflect what's intended to be implemented to a large extent. The details will be worked out by the contributors that take this CIP to its conclusion.

Thanks, that answers it: interpretation sits with the issuing and consuming apps, and an issuer that needs to withdraw on its own can run its own registry.

One detail for whoever takes the API forward, from the consuming side. In the draft OpenAPI, `listCredentials` returns the credentials and a page token, but not the ledger offset or record time the answer was computed at. An app that acts on a credential off-ledger, or compares reads from several Scan instances, needs that to state how old its reading is.

PR #209, opened 2026-05-13
`.canton` Naming CIP

dave-axymos:
Axymos launched xNS as a naming service on Canton Network in late 2025. 

Through the 'Identity and Metadata' SIG, we were invited to provide a CIP for how we could decentralise this as we move forward so that 

* multiple registrars could also offer names in the `.canton` namespace
* but all registrars are backed by a single shared canonical registry, to prevent fragmentation
* the service could outlive any individual company as needed

This is an early draft of the CIP and we'd welcome any questions/comments/suggestions.

### DIFF
--- a/cip-XXXX-canton-naming/cip-XXXX-canton-naming.md
+++ b/cip-XXXX-canton-naming/cip-XXXX-canton-naming.md
@@ -9,21 +9,41 @@
   Type: Standards Track
   Created: 2026-05-13
   License: CC0-1.0
+  Version: 1.1
 </pre>
+
+## Version History
+
+Updated to capture:
+* Clearer outline, focused on the end-user
+* Alignment to standards — `NameRecord` now implements CIP-56 token standard as well as the in-flight `RegistryV1.Credential`
+* `NameRegistry` now also implements `RegisterV1.CredentialFactory`
+* DRO is now an multi-hosted external party under a decentralised namespace definition, with each on-boarded registrar being 
+  * a quorum for the DND
+  * a signer of manual TXs as the DRO
+  * a whitelisted registrar in their own right (e.g. for attribution)
+* simplified elements like disputes on-chain (which can now be raised by a single registrar and then executed by simple DRO quorum)
+* simplified payment structure for things like after-market sales, which now go through CIP-56 standard flows for DvP.
+* simplified payment strucutre for registrations, which now are two straight transfers from…
+  * holder -> DRO
+  * holder -> facilitating registrar
+* added the concept of a "Treasury" contract for paying fees to DRO members, to reduce the number of transfers for each registration
+* added a "Pauseable" concept to the registry as an 'emergency break' on the system
+* Moved the concept of "reserved list" on-chain by pre-registering names that should be reserved at bootstrap phase (e.g. names of existing supervalidators/validators/featured apps, so they can be claimed by these parties)
 
 ## Abstract
 
 This CIP defines **Canton Naming**, a registry for human-readable `.canton` names on the Canton Network. Names like `alice.canton` resolve on-chain to Canton parties, providing a single source of truth for identity and discovery across decentralised applications.
 
-The protocol is designed up-front to be decentralised, where a pool of registrars can each sell name records to their end users but have them backed by a shared, logically centralised registry on-chain.
+The protocol is designed up-front to be decentralised, where a pool of registrars can each sell name records to their end users but have them backed by a shared, centralised registry on-chain.
 
 We believe that this is key to having a naming service be a value-add on the network, as names are only useful to end users if they can be relied upon to resolve to the same party. 
 
-The collective pool of registrars will operate a shared party, the "Decentralised Registry Operations party" (DRO, modelled on the DSO — the Decentralised Synchroniser Operations party that operates the Global Synchronizer). This will allow for all records to be created by a single party and means that we have a single maintainer of the contract keys used on-chain regardless of who has registered a given name.
+The collective pool of registrars will operate a shared party, the "Decentralised Registry Operator" (DRO, modelled on the DSO — the Decentralised Synchroniser Operator party that operates the Global Synchronizer). This will allow for all records to be created by a single party and means that we have a single maintainer of the contract keys used on-chain regardless of who has registered a given name.
 
 Governance of the service is managed by consensus among the approved registrars. Parameters of the service (like min pricing, vote thresholds etc) can be agreed upon by the governance layer and then are stored in the Name Registry contract itself. Registrars compete to provide name sales, renewals, and support to end users, but every name they sell is recorded in the same canonical `NameRegistry` contract. 
 
-The reference implementation (to follow) is a DAML contract package targeting Canton SDK 3.6.0. All authorisation flows through DAML's signatory model; the DAR would be vetted, so holders can exercise their own choices directly via the JSON Ledger API.
+The reference implementation (to follow) is a DAML contract package targeting Canton SDK 3.5.2. All authorisation flows through DAML's signatory model; the DAR would be vetted, so holders can exercise their own choices directly via the JSON Ledger API.
 
 ## Motivation
 
@@ -32,6 +52,110 @@ Applications building on the Canton Network today address users by raw Canton pa
 A naming layer only adds value if every name resolves to a single, agreed-upon entity — like a phone number or a bank account number. Two competing registries for the same namespace don't add resilience; they create confusion and erode user trust. Canton Naming is a **single source of truth** for `.canton` names: publicly queryable on-chain, backed by one canonical on-chain registry, while permitting any number of competing registrar products to participate in name sales, renewals, and support.
 
 Resilience is built in at both layers — the registry is operated by a multi-hosted DRO party so it survives any single hosting participant going down, and the set of registrars is governed on-chain so individual registrars can come and go without interrupting the registry itself.
+
+## Overview
+
+A higher-level walk-through of the system — the UX, the economic model, and the technical implementation — ahead of the formal specification below.
+
+### User flows
+
+- **Register a name** — signed by a facilitating registrar and the end-user holder. Users transfer funds from their wallet directly. The facilitating registrar takes a fee, with a remaining percentage going to the treasury (which provides a share of the fee to all whitelisted registrars, to cover the cost of governance operations). Name records are created as their own contract which itself implements CIP-56 and RegistryV1.Credentials interfaces
+- **Renew** — `UpdateCredentials` extends an existing record, signed by the holder and the DRO.
+- **Transfer** — supported via the CIP-56 standard.
+- **Sale** — registrars can facilitate a sale, which again leans on the CIP-56 interfaces.
+- **Expired names** — registrars can archive an expired record and reissue it to a new holder. This is not gated — any whitelisted registrar can do this. There's no vendor lock-in.
+
+### Governance flows
+
+There are a couple of "emergency brakes" on the system:
+
+- A whitelisted registrar can raise a dispute against a record. A disputed record can be archived by a k-of-n signing from the DRO, which constitutes a vote.
+- Any registrar can pause the registry if needed. Pausing freezes the registry's whole write surface — new registrations, transfers, allocations, and renewals. Unpausing requires a 2/3 registrar governance vote (`GA_Unpause`).
+
+Abuse of these is left to governance policy, agreed between the registrars themselves (TBD). They should be treated like "break glass" mechanisms — not expected to be used in normal operation of the system.
+
+Rogue registrars, acting in bad faith, can be removed: as a host of the DRO via the standard decentralised namespace definition flows, and from the on-chain registrar allowlist via a `GA_RemoveRegistrar` governance vote.
+
+To provide for fair competition between different companies, a baseline floor for pricing is set in the `NameRegistry` contract itself, and is subject to a registrar governance vote (`GA_UpdateFees`) to change.
+
+### UX
+
+*What problems are solved, and what UX flows are provided to solve them.*
+
+This CIP is aimed to be a shared foundation that allows multiple registrars to operate without conflict in the same shared namespace.
+
+In that light, the exact end-user experience that will be created in an individual registrar's dApp is out of scope of the CIP itself, and for competing offerings to decide upon. But each registrar should be able to offer the following:
+
+* Browse the registry and see which names are pre-existing
+* If a name isn't listed, sell it as available
+* Allow holders of names to renew, extending their expiry
+* Allow holders of names to sell or transfer the name to others
+
+ALl of these flows are unlocked "out of the box" by adhering to the existing CIP-56 standards and the in-flight standards from CIP-XXX (metadata/credentials) and would be supported by any registrar. There's no "vendor lock-in", i.e. a name registered by one registrar could be renewed by another etc.
+
+### Economic model
+
+*Who does what, and why are they willing to do this?*
+
+Individual companies can be onboarded as a registrar. Since they're hosting the DRO, we assume for simplicity that these would also be companies running their own validator to host the party. To onboard, they will:
+
+- Host the DRO
+- Have their own registrar party added as an approved (whitelisted) registrar
+
+The registrar can then facilitate registration / renewal / sale for end users. Facilitating registrars can take X percent of the fee for registration — they're competing on the UX of their product itself.
+
+Other registrars (since they're required to run the DRO) also get a fee share on registrations. The flow:
+
+- The user makes a payment
+- `NameRegistry` guarantees it's above the minimum floor and done on the basis of validaity (e.g. `minFloorRate * expiry`)
+- The fee that the end-user pays is split into two transfers, both sent by the user: the facilitating registrar's commission goes straight to them, and the remainder goes to the DRO treasury
+
+The treasury accrual to the registrar group can be withdrawn every X via a `Treasury` contract, which gives an even split of the treasury pool to currently whitelisted registrars.
+
+### Technical implementation
+
+*Component view: dApp(s), backends running at registrars, `.dar` packages vetted by users, `.dar` packages vetted by registrars only.*
+
+Since `NameRecords` are implementing `RegistryV1.Credential` the holder is a party on the contract instance itself. As such, we assume that any validator who wants to offer name sales to their user will have the DAR vetted. 
+
+Each whitelisted registrar would be expected to provide off-chain resolution of names from their own stack, but should be free to put controls in place to prevent abuse (e.g. API keys, rate limits etc).
+
+In addition to vetting the DAR, validators who are themselves registrars would also need to run a service to counter-sign transactions as the DRO, where needed.
+
+The "human in the loop" decisions are sparse (full table below) but mainly voting on disputes and on governance proposals (add/remove registrar, fee changes, unpausing) — routine registrations and renewals are signed automatically.
+
+End-users should have visibility over their own assets via the CIP-56 standard, even without the DAR installed, but interacting with their records would require the DAR available.
+
+| Role | Vets the XNS DAR | Holds a DRO key + runs signer/policy | Runs the registry HTTP service | In the on-ledger allowlist |
+|------|:---:|:---:|:---:|:---:|
+| **Registrar validator** | ✓ | ✓ | ✓ (may be shared) | ✓ |
+| **Holder validator** | ✓ | — | — | — |
+
+
+#### Information flows
+
+TBC.
+
+#### Key design decisions
+
+**Ensuring uniqueness of names.** There are two layers to this:
+
+1. How to stop a rogue registrar acting unilaterally
+2. How to prevent honest registrars from hitting a race condition
+
+For the first, `NameRecord`s require the DRO's signing authority. On bootstrapping of the service, this comes from a k-of-n ceremony between the initial hosting registrars. Any `NameRecord` created via `NameRegistry.RegisterName()` then inherits this automatically.
+
+A rogue registrar cannot bypass or forge this authority by manually creating a `NameRecord` — directly creating a DRO-signed contract needs fresh DRO authority, and an honest quorum of signers will refuse to countersign a bare record create. (The residual risk is a colluding quorum of k registrars.)
+
+For race conditions, we've implemented a sharding mechanism, whereby a "consistent hashing" of names picks a deployed shard to consume. In a race where honest registrars A and B both try to register `alice.canton` against the same offset:
+
+- They both prepare and submit a transaction that aims to consume the same shard
+- The loser of the race has their transaction rejected — it's trying to exercise a choice on a now-archived contract
+- Resubmitting at the next block means the contract keys now pick up the existing `alice.canton` record
+
+#### Link to PoC implementation
+
+TBC.
 
 ## Specification
 
@@ -51,12 +175,13 @@ On-chain: RegisterName (on NameRegistry)
   -> assertMsg "Invalid name format" (isValidName — lowercase, .canton suffix, no leading/trailing hyphens, 1–63 chars)
   -> assertMsg "Payment >= minPriceFloor"
   -> lookupByKey @NameRecord -> assertMsg "Name not already registered"
-  -> Fee split via chained TransferFactory calls:
-       1. Treasury transfer (paymentHoldingCids -> DRO); capture senderChangeCids
-       2. Sibling[0] transfer (senderChangeCids from step 1); capture senderChangeCids
-       3. Sibling[1] transfer (senderChangeCids from step 2); ...
-       Each transfer consumes its inputHoldingCids and returns new change holdings.
-       Registrar retains their fee as the final change holdings (no transfer needed).
+  -> Fee collection via two holder-funded TransferFactory_Transfer legs:
+       1. holder -> registrar, the registrar commission.
+       2. holder -> DRO treasury, the remainder.
+       The holder is the transfer sender and fee payer; the second leg
+       chains the first leg's change holdings so the unspent balance flows
+       through. The treasury remainder is later split across the registrars
+       by the Treasury contract's Treasury_Payout.
   -> create NameRecord (live immediately; usable from this point)
 ```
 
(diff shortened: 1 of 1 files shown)


cip-XXXX-canton-naming/cip-XXXX-canton-naming.md
```diff
@@ -0,0 +1,414 @@
...
+  Type: Standards Track
+  Created: 2026-05-13
+  License: CC0-1.0
+</pre>
+
+## Abstract
+
+This CIP defines **Canton Naming**, a registry for human-readable `.canton` names on the Canton Network. Names like `alice.canton` resolve on-chain to Canton parties, providing a single source of truth for identity and discovery across decentralised applications.
+
+The protocol is designed up-front to be decentralised, where a pool of registrars can each sell name records to their end users but have them backed by a shared, centralised registry on-chain.
```
meiersi-da, line 38:
```suggestion
The protocol is designed up-front to be decentralised, where a pool of registrars can each sell name records to their end users but have them backed by a shared, logically centralised registry on-chain.
```

cip-XXXX-canton-naming/cip-XXXX-canton-naming.md
```diff
@@ -0,0 +1,414 @@
...
+
+We believe that this is key to having a naming service be a value-add on the network, as names are only useful to end users if they can be relied upon to resolve to the same party. 
+
+The collective pool of registrars will operate a shared party, the "Decentralised Registry Operator" (DRO, modelled on the DSO — the Decentralised Synchroniser Operator party that operates the Global Synchronizer). This will allow for all records to be created by a single party and means that we have a single maintainer of the contract keys used on-chain regardless of who has registered a given name.
```
meiersi-da, line 42:
```suggestion
The collective pool of registrars will operate a shared party, the "Decentralised Registry Operations party" (DRO, modelled on the DSO — the Decentralised Synchroniser Operations party that operates the Global Synchronizer). This will allow for all records to be created by a single party and means that we have a single maintainer of the contract keys used on-chain regardless of who has registered a given name.
```

cip-XXXX-canton-naming/cip-XXXX-canton-naming.md
```diff
@@ -0,0 +1,414 @@
...
+
+Governance of the service is managed by consensus among the approved registrars. Parameters of the service (like min pricing, vote thresholds etc) can be agreed upon by the governance layer and then are stored in the Name Registry contract itself. Registrars compete to provide name sales, renewals, and support to end users, but every name they sell is recorded in the same canonical `NameRegistry` contract. 
+
+The reference implementation (to follow) is a DAML contract package targeting Canton SDK 3.6.0. All authorisation flows through DAML's signatory model; the DAR would be vetted, so holders can exercise their own choices directly via the JSON Ledger API.
```
meiersi-da: Why 3.6, that is not yet out. Did you mean 3.5, which is what provides contract keys?

cip-XXXX-canton-naming/cip-XXXX-canton-naming.md
```diff
@@ -0,0 +1,414 @@
...
+```
+Off-chain: Registrar does 'pre-flight' checking of availability / blacklist, holder funds etc.
+    |
+On-chain: RegisterName (on NameRegistry)
+  -> assertMsg "Registrar not in allowlist"
+  -> assertMsg "Invalid name format" (isValidName — lowercase, .canton suffix, no leading/trailing hyphens, 1–63 chars)
+  -> assertMsg "Payment >= minPriceFloor"
+  -> lookupByKey @NameRecord -> assertMsg "Name not already registered"
+  -> Fee split via chained TransferFactory calls:
+       1. Treasury transfer (paymentHoldingCids -> DRO); capture senderChangeCids
```
meiersi-da: We generally avoid having the DSO receiver transfers. We want it to only facilitate the flows, but not partake in the flows.

cip-XXXX-canton-naming/cip-XXXX-canton-naming.md
```diff
@@ -0,0 +1,414 @@
...
+       2. Sibling[0] transfer (senderChangeCids from step 1); capture senderChangeCids
+       3. Sibling[1] transfer (senderChangeCids from step 2); ...
+       Each transfer consumes its inputHoldingCids and returns new change holdings.
+       Registrar retains their fee as the final change holdings (no transfer needed).
+  -> create NameRecord (live immediately; usable from this point)
```
meiersi-da, line 184: It's not clear to me who pays whom and how much.

cip-XXXX-canton-naming/cip-XXXX-canton-naming.md
```diff
@@ -0,0 +1,414 @@
...
+```
+
+#### Transfer
+
+Names can be transferred either in the case of:
```
meiersi-da, line 189: Interesting idea. In that case, I'd suggest to represent names as CIP-56 assets by making each name a holding with `InstrumentId with admin = dso; id = 'name'` and `amount = 1.0`.

This way these names can be held and managed using a token standard wallet.

meiersi-da: And they can be traded.

cip-XXXX-canton-naming/cip-XXXX-canton-naming.md
```diff
@@ -0,0 +1,414 @@
...
+  -> assertMsg "Registrar in allowlist"
+  -> assertMsg "Name has expired" (expiresAt < now)
+  -> assertMsg "Has active disputes" (null disputes)
+  -> assertMsg "New expiry in future", "New expiry <= maxExtension"
+  -> Archive old -> Create new NameRecord
+```
+
+#### Dispute lifecycle
+
+Disputes occur in the case that a disputing registrar claims the registration should not stand — e.g. a rogue registrar registering an agreed reserved name, or a holder using the name in a way that violates registrar conduct policy.
```
meiersi-da: What defines an "agreed reserved name"?

cip-XXXX-canton-naming/cip-XXXX-canton-naming.md
```diff
@@ -0,0 +1,414 @@
...
+      |
+3. Counter-stake window:
+   a. No counter-stake by deadline
+      -> ClaimTimeout -> DisputeWon (disputer's stake returned)
+      -> jump to step 6
+   b. Counter-stake placed by any registrar in the allowlist
+      -> CounterStake (takes registryCid) -> fetches live registry -> sets voteDeadline (now + registry.voteWindow)
+      -> continue to step 4
+      |
+4. Registrars vote via AddVote (True = for dispute, False = against)
```
meiersi-da: This sounds like it might be quite a bit of work. What's the motivation for registrars to partake in these votes?

cip-XXXX-canton-naming/cip-XXXX-canton-naming.md
```diff
@@ -0,0 +1,414 @@
...
+#### Registration
+
+```
+Off-chain: Registrar does 'pre-flight' checking of availability / blacklist, holder funds etc.
+    |
+On-chain: RegisterName (on NameRegistry)
+  -> assertMsg "Registrar not in allowlist"
+  -> assertMsg "Invalid name format" (isValidName — lowercase, .canton suffix, no leading/trailing hyphens, 1–63 chars)
+  -> assertMsg "Payment >= minPriceFloor"
+  -> lookupByKey @NameRecord -> assertMsg "Name not already registered"
```
meiersi-da, line 176: There is no uniqueness guarantee for contract keys in Canton 3.x

I suspect that name allocation will have to be funnelled through an on-ledger representation of the map of all names (e.g. as a prefix tree) to ensure uniqueness even when there are many different nodes submitting name allocation transactions.

cip-XXXX-canton-naming/cip-XXXX-canton-naming.md
```diff
@@ -0,0 +1,414 @@
...
+
+2. **Registrar incentives** — Registrars are rewarded for their work via fees claimable from each registration and renewal. The fee split (registrar, sibling registrars, treasury) is governance-configurable and enforced on-chain via chained `TransferFactory_Transfer` calls.
+
+3. **Dispute resolution** — Generally new registrations should go through without issue as individual registrars are following the same standards and rules, but we've provided a dispute mechanism that registrars can use in the event of issues. This is deliberately light-weight in that it doesn't affect the "happy path" of registration (as disputes should be rare).
+
+4. **On-chain integrity** — Names are guaranteed unique via on-chain checks. We rely on the underlying infrastructure to solve for the race conditions/atomicity of these registrations. All authorisation flows through DAML's signatory model.
+
+
+### Open Questions
```
meiersi-da, line 541: I'm also missing clarity on the end-to-end flow for end-user actions: how will users build and submit their transactions (e.g., will they use a ".canton dApp" built as part of this CIP, or will they use their token standard wallet)? what transactions are built and submitted by end-users, which ones by registrars?

meiersi-da (commented):
Thanks @dave-axymos . This looks like a good first step. The current write-up is rather technical, and requires quite a few things to "reverse engineered".

What do you think about structuring the document to clearly separate:
1. the UX to be built: what problems are solved, and what UX flows are provided to solve them
2. the economic model: who does what and why are they willing to do this?
3. the technical implementation: split into
   1. component view: dApp(s), backends running at registrars, .dar packages vetted by users, .dar packages vetted by registrars only
   2. information flows: how does the information flow across the components to for the different UX flows described in (1)
   3. key design decisions: e.g., how to ensure uniqueness of names
   4. link to PoC implementation (where available)

dave-axymos:
> Thanks @dave-axymos . This looks like a good first step. The current write-up is rather technical, and requires quite a few things to "reverse engineered".
> 
> What do you think about structuring the document to clearly separate:
> 
>     1. the UX to be built: what problems are solved, and what UX flows are provided to solve them
> 
>     2. the economic model: who does what and why are they willing to do this?
> 
>     3. the technical implementation: split into
>        
>        1. component view: dApp(s), backends running at registrars, .dar packages vetted by users, .dar packages vetted by registrars only
>        2. information flows: how does the information flow across the components to for the different UX flows described in (1)
>        3. key design decisions: e.g., how to ensure uniqueness of names
>        4. link to PoC implementation (where available)

that sounds good, thanks @meiersi-da will restructure along those lines!

cip-XXXX-canton-naming/cip-XXXX-canton-naming.md
dave-axymos: The way we were looking to leverage contract keys was to do make sure that we gate our look-up on the fact that the multi-hosted party (the 'DRO') is the maintainer of the key and then it's choosing not to create duplicates.

So having a flow of:

1. Registrar e.g. Axymos does a look-up on chain to see if a name is available
2. If it is, calls `NameRegistry.RegisterName()` - which does a look-up of all active contracts owned collectively by the registrar pool and checks that it doesn't exist
3. So then using `len(results) == 0` as a proxy check for uniqueness.
4. Since the DRO is the key `maintainer` and since all registrars are registering as the DRO then our thoughts were that were safe enough to gate it

Couple of thoughts:

* we were leaning on the atomicity and sequencing of transactions on the network to make sure that race conditions prevented the second registration
* in the case of a "rogue registrar" (e.g. someone offboarded from the allowlist but still a host of the DRO) could bypass the path of driving all registrations through `RegisterName()` and instead just hit `createCmd` on NameRecord directly.

One thing we had that I seem to have accidentally dropped from `NameRecord` was the concept of activation. So thinking that the flow would be:

1. registrar calls `NameRegistry.RegisterName()`
2. registrats calls `NameRecord.Activate()`

Then resolving a name requires that it has been activated. I'm coming from a trustless EVM world so maybe that's overkill and we've better practices in Canton Network around stuff like this (e.g. literal contractual sign-up, say)

cip-XXXX-canton-naming/cip-XXXX-canton-naming.md
dave-axymos: Yeah this needs to be nailed down and better explaiend. 

Our thinking at the moment was that if there's some off-chain work.co-ordination needed between registrars then there would be fee split going on to account for that level of effort.

Then we'd also on-chain enforce things like a floor for prices

cip-XXXX-canton-naming/cip-XXXX-canton-naming.md
dave-axymos: that sounds really good - I was toying with this before and I can't remember why we moved away from it at the time. I'll revisit it in the reference impl to trial. thanks!

cip-XXXX-canton-naming/cip-XXXX-canton-naming.md
dave-axymos: > What defines an "agreed reserved name"?

At the moment for v1 live on chain, we've a reserve list that we don't let people register out of the gate — mainly profanity and clear phishing attempts e.g. stopping end users signing up as `hsbc.canton`, just to allow a decision to be made on that approach.

I think there's multiple options we could explore here:

- free rein — any name goes
- literally register the names that we think may want to be held (e.g. the names of each validator/supervalidator say) with a long expiry
- an off-chain list (which we have here at the moment)

I think we need to go away and strengthen up here the "what" and the "why" both and a path to someone claiming a reservered name etc if that's the why we go down.

cip-XXXX-canton-naming/cip-XXXX-canton-naming.md
dave-axymos: Appreciate the challenge to reframe here! :)

Again this is probably something that we're porting over from a trustless model. We're thinking of a flow similar to a Token Curated Registry where e.g.

1. Companies agree to stake X to become a registrar
2. In the event of a dispute, a contesting registrar can trigger a vote
3. There's a fee split between registrars voting
  a. each voting registrar receives financial incentive to vote
  b. winning registrar (disputer or challenger) receives a fee

In fact, based on the "do we need a reserve list" above, maybe this piece could naturally fall away - if there are no reserved names then there's no need to dispute on chain

cip-XXXX-canton-naming/cip-XXXX-canton-naming.md
dave-axymos: So existing in production for us, Axymos as registrar are submitting all TXs on the user behalf (as our DAR is running on our node, solely) - we were trying to move to a position where users would be able to do more with their own assets, so were thinking if this becomes a CIP standard, that 

- end-users would be able to have the DAR vetted on their node, so could interact directly, signing TXs as themselves
- hosting nodes would be able to query the `NameRegistry` "natively" from their own node (if added as an observer to the registry, or maybe if the DSO is a singleton observer? Note sure if that puts a burden on the DSO though)
- non-hosting nodes or external companies could still use APIs provided by any registrar to query the records.

At the moment, we're thinking that end users would co-sign TXs for things that they particular are actioning like if they choose to release a name (`Credential_ArchiveAsHolder()`) or for `TransferWithApproval` (to support transfer/sales/etc)

You're right that we've more focused the CIP on the on-chain elements that would allow registrars to build a full end-to-end product on top, rather than including elements like what a full dApp would look like.

We can definitely add more colour here to give an example, though we imagine that elements of UX etc will be driven by individual registrars building on top of this. But definitely will try to make the "bones" more obvious of who is signing what when, from where etc.

And I think if we do implement elements like CIP-56 then maybe we get more "for free" from the core standards (we're currently modelling it based on the credential inferface but maybe Token is more of a fit)

cip-XXXX-canton-naming/cip-XXXX-canton-naming.md
```diff
@@ -0,0 +1,414 @@
...
+- attack vectors of a "rogue" DRO host
+  - i.e. the ability to act unilaterally as one of the hosts of the multi-hosted party. (it's our understanding that a multi-hosted party is giving you technical failover but that there's no "human in the loop" element agreeing or declining to sign individual transactions) 
+
+## Backwards Compatibility
+
+No direct prior on-chain state for compatibility as an applications-layer CIP.
+
+## Reference Implementation
+
+Reference implementation to follow. Currently targeting Canton SDK 3.6.0 (LF target 2.3-staging), which is still in an alpha phase.
```
dave-axymos: We're working against contract keys, still in alpha, so are only on Sandbox at the moment, and we're also aiming to align with standards coming down the track like the RegistryV1.Credential, so have that stubbed out.

Other than that though, everything is based on working DAML code (though harder to test some aspects on a single node so sure there's more edge cases to find!) 

But I'll try to make time to tidy our reference implementation and get into a shareable state soon.

cip-XXXX-canton-naming/cip-XXXX-canton-naming.md
dave-axymos: I remember hitting an issue when trying to compile against 3.5 — I'll compile against that target again and double-check!

dave-axymos: btw, if 3.5 provides contract keys does that mean that they're available now on devnet @meiersi-da ?

cip-XXXX-canton-naming/cip-XXXX-canton-naming.md
meiersi-da: > And I think if we do implement elements like CIP-56 then maybe we get more "for free" from the core standards (we're currently modelling it based on the credential inferface but maybe Token is more of a fit)

I'd have expected for the `NameRecord` contracts to implement both the interfaces from the token standard and the ones from the credential standard. They server different purposes and enable different kinds of usages.

It's totally fine for one contract to implement multiple interfaces.

cip-XXXX-canton-naming/cip-XXXX-canton-naming.md
meiersi-da: Contract keys require connecting to a synchronizer running Canton Protocol Version 35 (PV=35), which in turn requires a Canton 3.5.x participant node that understands PV=35. For local development that's easy, just spin up a synchronizer with with PV=35 and your done. 

For DevNet this requires a logical synchronizer migration to change the synchchronizers protocol from its current PV=34 to PV=35. This is coming soon. Here's the CIP proposing the upgrade: https://github.com/canton-foundation/cips/pull/208/changes#diff-2fd1b7daef2c45396290d28ec49523ef503f8af706f2308ab1932751ffe1aea2

meiersi-da: And here's the operations schedule for the networks: https://docs.google.com/document/d/1QhLL5bL0u8temBL86y957VbWDtZJhH9udH-_C7nBlvc/edit?tab=t.0#heading=h.ripdn5ydglli

search for LSU

dave-axymos: @meiersi-da thanks, when I saw `0.6.4` drop yesterday I was hopefully it was on devnet already! 😅 

I've started to deploy a "scratchnet" super validator stack to EC2 just so I can have a play around with it from our own boxes as multiple validators. Will be great when it's on devnet! thanks for sharing the schedule, that's really helpful.

cip-XXXX-canton-naming/cip-XXXX-canton-naming.md
dave-axymos: Sorry I never replied to this — so this is our "DRO" here rather than the DSO proper. Did you mean that that's a pattern to avoid here too?

cip-XXXX-canton-naming/cip-XXXX-canton-naming.md
dave-axymos: > > And I think if we do implement elements like CIP-56 then maybe we get more "for free" from the core standards (we're currently modelling it based on the credential inferface but maybe Token is more of a fit)
> 
> I'd have expected for the `NameRecord` contracts to implement both the interfaces from the token standard and the ones from the credential standard. They server different purposes and enable different kinds of usages.
> 
> It's totally fine for one contract to implement multiple interfaces.

Thanks @meiersi-da, yeah I've updated our latest approach to implement both CIP-56 and the in-flight credential standard. I've also looked at moving `NameRegistry` itself to implement the credential factory (mapping the "renew" flow into `UpdateCredential`, but then keeping other flows like sales outside of the credential interface and leaning on the CIP-56 transferallocation and DvP flows).

dave-axymos:
@meiersi-da (and everyone!) since your feedback, have been working to:

- move the DRO to a proper externally hosted party — this simplifies a lot of flows, like e.g. the dispute flow is now:
  - one registrar raises a concern. Others vote on it via choosing to sign as the DRO and then the `k-of-n` threshold itself is what executes
- aligned to both CIP-56 and `RegistryV1.Credential` standards. I've created a little stub DAR of the credential interfaces to test out.
  -  reworked our transfer flows for sales around the CIP-56 standard on DvP.
  - for the CIP-56, I don't _think_ I can test this on localnet though, because the OOTB wallet UI doesn't seem to support custom tokens, is that right??
- implemented the `RegistryShard` concept so that
  -  honest path race conditions are avoided by the contract archiving
  - rogue paths are avoided because
    - single host of the DRO can't act with its authority solo as it's now properly an external party so needs `k-of-n` quorum
    - honest `RegisterName()` calls don't pay the "penalty" of counter-signing, since the authority flows through the canonical `NameRegistry` contract, which is bootstrapped at the start of the network

I'm going to try and update this draft to align with these changes now, but also trying to get a scratchnet set-up working so that can test more of these corner cases with real code before locking in on shapes

meiersi-da:
> @meiersi-da (and everyone!) since your feedback, have been working to:
> 
> * move the DRO to a proper externally hosted party — this simplifies a lot of flows, like e.g. the dispute flow is now:

Note: we usually distinguish between decentralized parties that cannot submit transactions on their own and external parties whose tx signing keys are held externally to the validator nodes. The DSO party used for CC is setup as a decentralized party using a decentralized namespace.

>   * one registrar raises a concern. Others vote on it via choosing to sign as the DRO and then the `k-of-n` threshold itself is what executes
> * aligned to both CIP-56 and `RegistryV1.Credential` standards. I've created a little stub DAR of the credential interfaces to test out.
>   
>   * reworked our transfer flows for sales around the CIP-56 standard on DvP.
>   * for the CIP-56, I don't _think_ I can test this on localnet though, because the OOTB wallet UI doesn't seem to support custom tokens, is that right??

The Splice Amulet UI does not. However you should be able to run https://github.com/canton-network/wallet/tree/main/examples/portfolio to get a CIP-56 wallet UI. Haven't tried it myself yet, but heard good things.

dave-axymos:
> Note: we usually distinguish between decentralized parties that cannot submit transactions on their own and external parties whose tx signing keys are held externally to the validator nodes. The DSO party used for CC is setup as a decentralized party using a decentralized namespace.

thanks Simon, I think I understand the split as:   

* decentralised namespace definition - control over the DRO party itself (e.g. onboarding a host of the DRO, key rotation etc) 
* then `k-of-n` signing - evaluating and signing individual TXs.  

Is that accurate? 

If so, we’re looking to split at both layers, but initially thinking of modelling them 1-to-1 (so an onboarded registrar would become a quorum party to the DND and would hold a key for signing authority as the DRO). I believe this matches the DSO model.   

We’re also assuming (for simplicity) that each host of the DRO are themselves a validator so the “party to participant” is straightforward. (And we don’t have, say, a registrar who has their own external key but who isn’t running a validator and are hosted on elsewhere. Though guess in theory we could open this up, too if desired!)

  Happy to remodel that if you think a different take on this would be better for our use-case? (Or maybe in the future we decide that e.g. not every registrar needs to be part of the DND itself)   But wanted to start with a structure that felt open to collaboration.   
  
  I'm also going to try and put more of a breakdown in next draft of where authority: 
 
* a) flows from a pre-signed contract
* b) is a direct quorum of governance
* c) is something else e.g. 2/3s majority supported by a contract 

> The Splice Amulet UI does not. However you should be able to run https://github.com/canton-network/wallet/tree/main/examples/portfolio to get a CIP-56 wallet UI. Haven't tried it myself yet, but heard good things.

Ah cool thanks! Will look at getting wallet bridge set-up on localnet and see if I can hook that in too.

meiersi-da:
On decentralized party: consider building on https://github.com/gabitu7/canton-dev-fund/blob/decentralization-manager-proposal/proposals/decentralization-manager.md

cip-XXXX-canton-naming/cip-XXXX-canton-naming.md
```diff
@@ -0,0 +1,561 @@
...
+
+- Host the DRO
+- Have their own registrar party added as an approved (whitelisted) registrar
+
+The registrar can then facilitate registration / renewal for end users. Facilitating registrars can take X percent of the fee for registration — they're competing on the UX of their product itself.
+
+Other registrars (since they're required to run the DRO) also get a fee share on registrations. The flow:
+
+- The user makes a payment
+- `NameRegistry` guarantees it's above the minimum floor and done on the basis of validaity (e.g. `minFloorRate * expiry`)
```
leonidr-c7:
```suggestion
- `NameRegistry` guarantees it's above the minimum floor and done on the basis of validity (e.g. `minFloorRate * expiry`)
```

cip-XXXX-canton-naming/cip-XXXX-canton-naming.md
```diff
@@ -0,0 +1,561 @@
...
+A higher-level walk-through of the system — the UX, the economic model, and the technical implementation — ahead of the formal specification below.
+
+### User flows
+
+- **Register a name** — signed by a facilitating registrar and the end-user holder. Users transfer funds from their wallet directly. The facilitating registrar takes a fee, with a remaining percentage going to the treasury (which provides a share of the fee to all whitelisted registrars, to cover the cost of governance operations). Name records are created as their own contract which itself implements CIP-56 and RegistryV1.Credentials interfaces
+- **Renew** — `UpdateCredentials` extends an existing record, signed by the holder and the DRO.
+- **Transfer** — supported via the CIP-56 standard.
+- **Expired names** — registrars can archive an expired record and reissue it to a new holder. This is not gated — any whitelisted registrar can do this. There's no vendor lock-in.
+
+### Governance flows
```
luis-marado-unlockit, line 67: As I was reading this I linked to this proposal that is actively being worked on https://github.com/canton-foundation/canton-dev-fund/pull/184, may be relevant to coordinate actions cc @meiersi-da @dave-axymos

PR #212, opened 2026-05-17
Update CIP to incorporate latest iterations

bame-da:
This incorporates the changes communicated via https://lists.sync.global/g/cip-discuss/topic/118630150.

### DIFF
--- a/cip-0112/cip-0112.md
+++ b/cip-0112/cip-0112.md
@@ -119,7 +119,7 @@ This opens up using standard transfer and allocation workflows for mint and burn
 
 ### 4.1.4 Improvements to Timing Parameters
 
-In line with practices in traditional financial markets, the time-related fields within `SettlementInfo` are changed to allow more flexibility. The previously mandatory inclusion of `settleBefore` and `allocateBefore` deadlines is too restrictive for workflows where settlement is not driven by time but by external business events or bilateral agreements.
+In line with practices in traditional financial markets, the time-related fields within `SettlementInfo` are changed to allow more flexibility. The previously mandatory inclusion of `settleBefore` and `allocateBefore` deadlines is too restrictive for workflows where settlement is not driven by time but by external business events or bilateral agreements. The timing parameters have also moved form `SettlementInfo` to `AllocationSpecification`, and `SettlementInfo` has been turned into `SettlementReference` to more clearly encapsulate that part of the `AllocationSpecification` which links it to an overall settlement.
 
 ### 4.1.5 Improved Transaction Parsing
 
@@ -183,7 +183,7 @@ To illustrate the intended settlement flow, this section walks through a trade u
 
 ### 4.2.1 Setup
 
-Having executed a trade off-chain, the App app defines the `SettlementInfo` with `executors=[app]` and a list of nine trading legs:
+Having executed a trade off-chain, the App app defines the `SettlementReference` with `executors=[app]` and a list of nine trading legs:
 
 ```
 [
@@ -343,7 +343,7 @@ Transaction:
 
 ### 4.2.4 Settlement (TX6)
 
-The App now takes all 7 allocations and in the same transaction sticks them into new choices `SettlementFactory_SettleBatch` of the X and \$ `SettlementFactories` with just the App as `actor` as specified in the `SettlementInfo`:
+The App now takes all 7 allocations and in the same transaction sticks them into new choices `SettlementFactory_SettleBatch` of the X and \$ `SettlementFactories` with just the App as `actor` as specified in the `SettlementReference`:
    1. 3.a,c,e into the X `SettlementFactory`
    2. 4.b,d,f,g into the \$ `SettlementFactory`
 
@@ -381,7 +381,7 @@ In the current state, an Allocation covers exactly one transferLeg, and is settl
 
 ```
 data AllocationSpecification = AllocationSpecification with
-    settlement : SettlementInfo
+    settlement : SettlementReference
     transferLegId : Text
     transferLeg : TransferLeg
 
@@ -400,7 +400,7 @@ The trust model is that `[executor, sender, receiver]` jointly guarantee atomici
 
 #### *Proposed change*
 
-Traders (or specifically Trading Accounts after taking into account [4.3.2](#432-accounts-instead-of-parties)) should create a single `Allocation` per settlement and `admin` PartyId which covers all legs on assets managed by that `admin` Party where the trader is the `sender` or `receiver`. That way the `admin` Party can guarantee atomicity within their system. To allow this, while keeping data compact and disambiguating whether the `Allocation` authorizer is on the sending or receiving leg of a self-transfer, `AllocationSpecifications` now include the authorizing `Account` and contain a list of `transferLegSides`, each specifyign whether the authorizer is on the sending or receiving end. The `transferLeg` id is moved into the `TransferLegSide` type.
+Traders (or specifically Trading Accounts after taking into account [4.3.2](#432-accounts-instead-of-parties)) SHOULD (with carve outs in 4.3.8) create a single `Allocation` per settlement and `admin` PartyId which covers all legs on assets managed by that `admin` Party where the trader is the `sender` or `receiver`. That way the `admin` Party can guarantee atomicity within their system. To allow this, while keeping data compact and disambiguating whether the `Allocation` authorizer is on the sending or receiving leg of a self-transfer, `AllocationSpecifications` now include the authorizing `Account` and contain a list of `transferLegSides`, each specifyign whether the authorizer is on the sending or receiving end. The `transferLeg` id is moved into the `TransferLegSide` type.
 
 ```
 data TransferLeg = TransferLeg with
@@ -419,7 +419,7 @@ data TransferLegSide = TransferLegSide with
   ...
 
 data AllocationSpecification = AllocationSpecification with
-    settlement : SettlementInfo
+    settlement : SettlementReference
     transferLegSides : [TransferLegSide]
     authorizer : Account
       -- ^ The account authorizing the transfers to happen as part of the settlement.
@@ -428,7 +428,7 @@ data AllocationSpecification = AllocationSpecification with
 The `executor` field is now a list to allow for atomicity guarantees to be distributed across multiple parties, including the V1 default set `[executor, sender, receiver]`.
 
 ```
-data SettlementInfo = SettlementInfo
+data SettlementReference = SettlementReference
   with
     executors : [Party]
       -- ^ The parties that are responsible for executing the settlement and guarantee atomicity across admin Parties.
@@ -448,7 +448,7 @@ interface SettlementFactory where
     -- each account across all transfers being settled; and restrict visibility of
     -- each credit or debit to executors, admin, and affected account parties only.
     with
-      settlement : SettlementInfo
+      settlement : SettlementReference
         -- ^ The settlement for which the allocations are settled.
       transferLegs : [TransferLeg]
         -- ^ The transfers that are to be executed as part of the settlement.
@@ -618,7 +618,7 @@ Token implementations that allow the user to request burning or minting tokens a
 * account-id \= `cip-112/mint` for a mint
 * account-id \= `cip-112/burn` for a burn
 
-This guidance ensures that wallets can present a uniform user experience for mints and burns by interpreting the above account ids specially. This also allows apps to uniformly include burn and mint legs in `Allocation`-based settlement.
+This guidance ensures that wallets can present a uniform user experience for mints and burns by interpreting the above account ids specially. This also allows apps to uniformly include burn and mint legs in `Allocation`-based settlement and analogously to create burn requests as transfer-offers to the burn account, and mint-offers as transfer-offers from the mint account.
 
 #### 4.3.2.2 Implicit and Explicit Accounts and Account Discovery
 
@@ -647,7 +647,7 @@ For the avoidance of doubt, these fields have not been changed as they directly 
 
 * Any fields like `admin`, `actors`, etc.
 * `Lock.holders` remains of type `[Party]`
-* `SettlementInfo.executors` remains of type `[Party]`
+* `SettlementReference.executors` remains of type `[Party]`
 
 Specifically, the following choices now take the extra field `actors` of type `[Party]`:
 
@@ -663,7 +663,7 @@ Specifically, the following choices now take the extra field `actors` of type `[
 
 ### 4.3.3 Improved Timing Parametrization
 
-The fields `settleBefore` and `allocateBefore` are replaced on `SettlementInfo` by a mandatory `settleAt` and an optional `settlementDeadline`.
+The timing parameters `settleBefore` and `allocateBefore` on `SettlementInfo` have been reworked and moved to different types. `SettlementInfo` is renamed to `SettlementReference` to indicate more clearly that that's what should be matched on across `Allocations`.
 
 #### *Current State*
 
@@ -677,39 +677,82 @@ data SettlementInfo = SettlementInfo
 
 #### *Proposed Change*
 
+The `SettlementReference` type is reduced to this:
+
 ```
-data SettlementInfo = SettlementInfo
+-- | An unambiguous reference to a settlement, which can be used by wallets to
+-- correlate allocations to the same settlement.
+--
+-- The `executors` MUST ensure that the triple of `(id, cid, meta)` is unique
+-- across settlements.
+data SettlementReference = SettlementReference
   with
-    --{other args}
-    settleAt : Time
-      -- ^ The earliest settlement time. Allocations should be made before this time. Settlement
-      -- happens at any point after this time.
-    settlementDeadline : Optional Time
-      -- ^ The latest point at which settlement will occur. This allows registries
-      -- to set appropriate locks and expiries.
+    executors : [Party]
+      -- ^ The parties that are responsible for executing the settlement and
+      -- guarantee atomic settlement across asset admins.
+    id : Text
+      -- ^ Textual identifier of the settlement.
+    cid : Optional AnyContractId
+      -- ^ Optional contract-id based identifier of the settlement.
+      --
+      -- This field is there for technical reasons, as contract-ids cannot be
+      -- converted to text from within Daml, which is due to their full textual
+      -- representation being only known after transactions have been prepared.
+    meta : Metadata
+      -- ^ Additional metadata to identify the settlement, used for extensibility.
+  deriving (Show, Eq)
 ```
 
-As indicated by the code comment, `settlementDeadline` acts as a hint to wallets and registries that they should set any locks and expiries to accommodate settlement before this point in time, and may collect garbage thereafter.
+The timings are now spread across three places.
 
-Assets MAY have maximum durations for locks and allocations, or general timeouts to allow garbage collection of abandoned processes. To complement `settlementDeadline` on the SettlementInfo, `Allocations`, and `AllocationInstructions` now have a field `expiresAt` which informs apps and wallets when the contracts actually expire; and can thus no longer be settled, but can be tidied up.
+First, `AllocationRequest` has fields for the requester to state their desired timings. `requestedAt` is the time at which the app requested allocation, `settleAt` is an optional time at which the app expects to actually settle. Users and wallets should aim to allocate (or reject) before this time.
 
-#### *Proposed Change*
+```
+data AllocationRequestView = AllocationRequestView with
+    ...
+    requestedAt : Time
+      -- ^ Timestamp at which the request was created.
+    settleAt : Optional Time
+      -- ^ Timestamp at which the settlement is expected to be executed. The authorizer
+      -- SHOULD create their allocations before this time.
+      --
+      -- For iterated settlements, this is the expected time of the first iteration.
+```
+
+Second, on `AllocationSpecification`, the app can set an optional `settlementDeadline`, which the authorizer agrees to as part of creating an `Allocation`. Assets should ensure that after this time passes, `Allocation` can no longer be settled, and the `authorizer` can withdraw it, even if `committed`.
+
+```
+    settlementDeadline : Optional Time
+      -- ^ The executors' and authorizer's agreed time-to-live for the allocation.
+      --
+      -- After this time, if set, the allocation can no longer be settled, and
+      -- the authorizer can withdraw the allocation to release the funds.
+```
+
+And third, the asset itself can ensure that `Allocations` have a finite time to live by setting `expiresAt` on `AllocationView`.
 
 ```
 data AllocationView = AllocationView with
-    --{other args}
+    ...
     expiresAt : Optional Time
-    -- ^ A time at which this allocation expires. SHOULD be as close as the
-    -- registry can make it to the settlementInfo's settlementDeadline.
+      -- ^ The time at which the allocation expires if inactive.
+      --
+      -- Registries MAY expire the allocation and return the locked funds to the
+      -- authorizer after this time. Thereby recovering storage resources and
+      -- protecting themselves from denial-of-service attacks.
+      --
+      -- Registries SHOULD avoid unnecessary refreshes by
+      --   - making the expiry time as close to the settlement deadline as possible
+      --   - bumping expiry on every settlement iteration.
 ```
 
-Registries SHOULD set `expiresAt` to be equal to `settlementDeadline` if possible, or to the maximum value if not. Value `None` means there is no expiry and users/apps need to tidy up manually.
+Registries SHOULD set `expiresAt` to be equal to `settlementDeadline` if possible, or to the maximum value if not. Value `None` means there is no expiry and users/apps need to tidy up manually. 
 
 ## 4.3.5 EventLog for Transaction Parsing
 
 CIP-0112 improves on transaction parsing by decoupling the events needed to read transfers and holding changes in a standardized way from the rest of the logic. A new package `splice-api-token-transfer-events-v2` specifies an interface `EventLog` with a single nonconsuming choice `EventLog_HoldingsChange`. This choice SHOULD be implemented such that it has no side-effects. The choice's purpose is to make the call arguments visible to its observers, nothing more.
 
-At the core, the call arguments consist of the input- and output holding ContractIds, and the transfers that resulted in these changes. The ContractId deltas allow efficient updates of indices. `HoldingViews` are not included for space efficiency. The `HoldingViews` can be read from the actual create and archive events of the holdings in the same transaction, or retrieved from separate endpoints on the Ledger API.
+At the core, the call arguments consist of the input- and output holding `ContractIds`, and the transfers that resulted in these changes. The `ContractId` deltas allow efficient updates of indices. `HoldingViews` are not included for space efficiency. The `HoldingViews` can be read from the actual create and archive events of the holdings in the same transaction, or retrieved from separate endpoints on the Ledger API.
 
 ```
 -- Note: the TransferLegSide type below is identical to AllocationV2.TransferLegSide, but
@@ -886,7 +929,7 @@ interface SettlementFactory
         -- instrument holder.
 
         -- To minimize the risk of allocations being settled in the wrong
-        -- context, implementations SHOULD ensure that:
+        -- context, implementations MUST ensure that:
         --
         --  - the allocations were created for the `settlement` being settled
         --  - the transfer legs of the allocations authorize both send and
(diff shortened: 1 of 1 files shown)


cip-0112/cip-0112.md
```diff
@@ -109,18 +111,75 @@ Case in point, CIP-0056 `Allocations` which are core to the settlement flows rel
...
 
+### 4.1.3.1 Accounts representing Non-Holder Destinations
+
+Using Accounts rather than `Party` as the source and destination of operations like transfers also allows for the clean representation of non-transfer operations like burns and mints. This is analogous to Ethereum where commonly the `null` address (0x0000000000000000000000000000000000000000) is used as the source/destination for mints and burns. Canton doesn't have a "null Party", but the `Account` data structure can represent special ownerless accounts that represent the source and destination of mints and burns, respectively.
+
+This opens up using standard transfer and allocation workflows for mint and burn operations. Most notably, it opens up the possibility to settle "Delivery vs Mint" and "Delivery vs Burn" settlements, where fund or stablecoin operators can allow subscriptions or redemptions as part of trades.
```
meiersi-da:
```suggestion
This opens up using standard transfer and allocation workflows for mint and burn operations. Most notably, it opens up the possibility to settle "Delivery vs Mint" and "Delivery vs Burn" trades, where fund or stablecoin operators can allow subscriptions or redemptions as part of trades.
```

note sure

cip-0112/cip-0112.md
```diff
@@ -109,18 +111,75 @@ Case in point, CIP-0056 `Allocations` which are core to the settlement flows rel
...
+
 ### 4.1.4 Improvements to Timing Parameters
 
 In line with practices in traditional financial markets, the time-related fields within `SettlementInfo` are changed to allow more flexibility. The previously mandatory inclusion of `settleBefore` and `allocateBefore` deadlines is too restrictive for workflows where settlement is not driven by time but by external business events or bilateral agreements.
```
meiersi-da: We shifted `settleAt` to allocation requests and allocation views, and moved the `SettlementInfo.requestedAt` to `AllocationView.createdAt`.

Given our convo on `settlementDeadline`, I believe we should update this section to reflect the changes.

cip-0112/cip-0112.md
```diff
@@ -109,18 +111,75 @@ Case in point, CIP-0056 `Allocations` which are core to the settlement flows rel
...
+### 4.1.5 Improved Transaction Parsing
+
+CIP-0112 introduces a new `EventLog` interface, which emits standalone events as side-effect free non-consuming choices, similar to ERC-20 events. This simplifies parsing for wallets, exchanges and apps: just look for `EventLog_HoldingsChange` events to understand all changes to the holdings of an account.
+It also makes things easier for asset implementations as they get more fine grained control over who sees what.
+
+To contrast, [CIP-0056 transaction history parsing](https://docs.digitalasset.com/integrate/devnet/exchange-integration/txingestion.html#differences-between-1-step-deposits-and-withdrawals) relies on events from the `TransactionFactory` and `AllocationFactory` interfaces with a fallback to metadata. Both have some shortcomings: Using the factory events means that observing the event means observing all consequences thereof. E.g. to show a receiver of a transfer the `TransactionFactory_Transfer` event, one would also have to show them the input holding pre-split. Using metadata requires full transaction tree traversal on the client side rather than being able to rely on a list of filtered events.
```
meiersi-da:
```suggestion
To contrast, [CIP-0056 transaction history parsing](https://docs.digitalasset.com/integrate/devnet/exchange-integration/txingestion.html#differences-between-1-step-deposits-and-withdrawals) relies on interpreting the choices from the `TransactionFactory`, `TransactionInstruction`, `AllocationFactory`, `AllocationInstruction`, and `Allocation` interfaces as events with a fallback to metadata. Both have some shortcomings: Using the interfaces choices as events means that observing the event means observing all consequences thereof. E.g. to show a receiver of a transfer the `TransactionFactory_Transfer` event, one would also have to show them the input holding pre-split. Using metadata requires full transaction tree traversal on the client side rather than being able to rely on a list of filtered events.
```

cip-0112/cip-0112.md
```diff
@@ -109,18 +111,75 @@ Case in point, CIP-0056 `Allocations` which are core to the settlement flows rel
...
+CIP-0056's transaction parsing methodology is also a challenge for the cross-version compatibility to work fully on the transaction history-parsing side. If the same factory-choice based approach was adopted in CIP-0112, and one wanted to emit both V1 and V2 events for a transfer, then the V1 choice has to call the V2 choice and vice versa. This circular dependency with the needed switching logic whether the choice is called directly or by the other version is not practical.
+
+### 4.1.6 Committed Allocations for Prefunded Trading and Iterated Settlement
+
+`Allocations` in the token standard are intended to represet a a given amount of assets that is guaranteed to be settleable by the `executor` in the context of a broader trade. Atomic settlement is achieved through a four step process:
```
meiersi-da:
```suggestion
`Allocations` in the token standard V1 are intended to represet a a given amount of assets that is guaranteed to be settleable by the `executor` in the context of a broader trade. Atomic settlement is achieved through a four step process:
```

?

cip-0112/cip-0112.md
```diff
@@ -109,18 +111,75 @@ Case in point, CIP-0056 `Allocations` which are core to the settlement flows rel
...
+ - the `executors` cancel the `Allocation`,
+ - the settlement deadline passed, or
+ - the `admin` expires the `Allocation`.
+
+This gives an on-chain exchange like the above a choice:
+
+1. Allow the user to withdraw unilaterally, minimizing counterparty risk to the exchange, but risking races between trades and withdrawals.
+2. Asking traders to commit funds, in which case traders have to trust the exchange to honour withdrawals.
+
+This ability to grant committed control to a given pool of assets to an app is also a core primitive of many on-chain protocols including lending, and staking. In such cases, withdrawals would then typically be controlled by the app's smart contracts, giving rise to the exact patters common in DeFi. But instrad of transferring assets to the contract address of the protocol as is commen in EVM, the assets stays clearly owned by the trader, but is in a committed, allocated state to the protocol, allowing it to act on the asset according to the rules of the protocol.
```
meiersi-da:
```suggestion
This ability to grant committed control to a given pool of assets to an app is also a core primitive of many on-chain protocols including lending, and staking. In such cases, withdrawals would then typically be controlled by the app's smart contracts, giving rise to the exact patters common in DeFi. But instead of transferring assets to the contract address of the protocol as is commen in EVM, the assets stays clearly owned by the trader, but is in a committed, allocated state to the protocol, allowing it to act on the asset according to the rules of the protocol.
```

cip-0112/cip-0112.md
```diff
@@ -341,18 +400,27 @@ The trust model is that `[executor, sender, receiver]` jointly guarantee atomici
...
-Since an `Allocation` now contains legs where the Trading Account could be in sender or receiver position, the `AllocationSpecification` includes the authorizing `Account`.
+Traders (or specifically Trading Accounts after taking into account [4.3.2](#432-accounts-instead-of-parties)) should create a single `Allocation` per settlement and `admin` PartyId which covers all legs on assets managed by that `admin` Party where the trader is the `sender` or `receiver`. That way the `admin` Party can guarantee atomicity within their system. To allow this, while keeping data compact and disambiguating whether the `Allocation` authorizer is on the sending or receiving leg of a self-transfer, `AllocationSpecifications` now include the authorizing `Account` and contain a list of `transferLegSides`, each specifyign whether the authorizer is on the sending or receiving end. The `transferLeg` id is moved into the `TransferLegSide` type.
```
meiersi-da: I'm wondering whether "should create a single allocation" is still true with the limits on the number of legs. I'd expect them to create multiple allocations in some cases to circumvent the number of legs limitation.

It's a should, so probably its not worth mentioning here.

Independently: do you mention somewhere the limit on the number of legs, and the documented minimum?

meiersi-da:
```suggestion
Traders (or specifically Trading Accounts after taking into account [4.3.2](#432-accounts-instead-of-parties)) should create a single `Allocation` per settlement and `admin` PartyId which covers all legs on assets managed by that `admin` Party where the trader is the `sender` or `receiver`. That way the `admin` Party can guarantee atomicity within their system. To allow this, while keeping data compact and disambiguating whether the `Allocation` authorizer is on the sending or receiving leg of a self-transfer, `AllocationSpecification`s now include the authorizing `Account` and contain a list of `transferLegSides`, each specifyign whether the authorizer is on the sending or receiving end. The `transferLeg` id is moved into the `TransferLegSide` type.
```

cip-0112/cip-0112.md
```diff
@@ -535,14 +611,14 @@ choice TransferInstruction_Accept : TransferInstructionResult
...
+#### 4.3.2.1 Special Account Identifiers for Mint and Burn
 
-Token implementations that allow the user to request burning or minting tokens as part of an allocation flow SHOULD represent them as transfers to and from the `admin` party with the following account-ids:
+Token implementations that allow the user to request burning or minting tokens as part of an allocation flow SHOULD represent them as transfers to a special account with `owner` and `provider` equal to `None` and account-ids:
 
 * account-id \= `cip-112/mint` for a mint
 * account-id \= `cip-112/burn` for a burn
 
-This guidance ensures that wallets can present a uniform user experience for mints and burns by interpreting the above account ids specially.
+This guidance ensures that wallets can present a uniform user experience for mints and burns by interpreting the above account ids specially. This also allows apps to uniformly include burn and mint legs in `Allocation`-based settlement.
```
meiersi-da: and analogously to create burn requests as transfer-offers to the burn account, and mint-offers as transfer-offers from the mint account.

cip-0112/cip-0112.md
```diff
@@ -615,7 +691,7 @@ data SettlementInfo = SettlementInfo
 
 As indicated by the code comment, `settlementDeadline` acts as a hint to wallets and registries that they should set any locks and expiries to accommodate settlement before this point in time, and may collect garbage thereafter.
 
-Assets MAY have maximum durations for locks and allocations. To complement `settlementDeadline` on the SettlementInfo, `Allocations` now have a field `expiresAt` which informs apps and wallets when the `Allocation` actually expires and can no longer be settled.
+Assets MAY have maximum durations for locks and allocations, or general timeouts to allow garbage collection of abandoned processes. To complement `settlementDeadline` on the SettlementInfo, `Allocations`, and `AllocationInstructions` now have a field `expiresAt` which informs apps and wallets when the contracts actually expire, can no longer be settled, but can be tidied up.
```
meiersi-da:
```suggestion
Assets MAY have maximum durations for locks and allocations, or general timeouts to allow garbage collection of abandoned processes. To complement `settlementDeadline` on the SettlementInfo, `Allocations`, and `AllocationInstructions` now have a field `expiresAt` which informs apps and wallets when the contracts actually expire; and can thus no longer be settled, but can be tidied up.
```

cip-0112/cip-0112.md
```diff
@@ -629,36 +705,314 @@ data AllocationView = AllocationView with
 
 Registries SHOULD set `expiresAt` to be equal to `settlementDeadline` if possible, or to the maximum value if not. Value `None` means there is no expiry and users/apps need to tidy up manually.
 
-## 4.3.4 App, Wallet, and Asset Guidelines for Allocations
+## 4.3.5 EventLog for Transaction Parsing
+
+CIP-0112 improves on transaction parsing by decoupling the events needed to read a transaction in a standardized way from the rest of the logic. A new package `splice-api-token-transfer-events-v2` specifies an interface `EventLog` with a single nonconsuming choice `EventLog_HoldingsChange`. This choice SHOULD be implemented such that it has no side-effects. The choice's purpose is to make the call arguments visible to its observers, nothing more.
```
meiersi-da:
```suggestion
CIP-0112 improves on transaction parsing by decoupling the events needed to read transfers and holding changes in a standardized way from the rest of the logic. A new package `splice-api-token-transfer-events-v2` specifies an interface `EventLog` with a single nonconsuming choice `EventLog_HoldingsChange`. This choice SHOULD be implemented such that it has no side-effects. The choice's purpose is to make the call arguments visible to its observers, nothing more.
```

cip-0112/cip-0112.md
```diff
@@ -629,36 +705,314 @@ data AllocationView = AllocationView with
...
+
+At the core, the call arguments consist of the input- and output holding ContractIds, and the transfers that resulted in these changes. The ContractId deltas allow efficient updates of indices. `HoldingViews` are not included for space efficiency. The `HoldingViews` can be read from the actual create and archive events of the holdings in the same transaction, or retrieved from separate endpoints on the Ledger API.
```
meiersi-da:
```suggestion
At the core, the call arguments consist of the input- and output holding `ContractId`s, and the transfers that resulted in these changes. The `ContractId` deltas allow efficient updates of indices. `HoldingViews` are not included for space efficiency. The `HoldingViews` can be read from the actual create and archive events of the holdings in the same transaction, or retrieved from separate endpoints on the Ledger API.
```

cip-0112/cip-0112.md
```diff
@@ -629,36 +705,314 @@ data AllocationView = AllocationView with
...
+
+Registries MAY allow the creations of `Allocations` with nonempty `transferLegSides` _and_ `nextIterationFunding` at the same time. In that case, the legs in `transferLegSides` are guaranteed to be settled as part of the first iteration, after which `transferLegSides` is always empty.
+
+The `executors` specify transfer legs for settlement as input to `SettlementFactory_SettleBatch`, using the `extraTransferLegSides` field of the `FinalizedAllocation` type. There, they can also specify the new `nextIterationFunding` for the next iteration. If set to `None`, all funds resulting from the settlement are returned to the user. If set to a map containing quantities not exceeding those available after settlement, a new `Allocation` is output with the specified `nextIterationFunding`. Any excess funds are returned to the `authorizer`.
```
meiersi-da:
```suggestion
The `executors` specify transfer legs for settlement as input to `SettlementFactory_SettleBatch`, using the `extraTransferLegSides` field of the `FinalizedAllocation` type shown below. There, they can also specify the new `nextIterationFunding` for the next iteration. If set to `None`, all funds resulting from the settlement are returned to the user. If set to a map containing quantities not exceeding those available after settlement, a new `Allocation` is output with the specified `nextIterationFunding`. Any excess funds are returned to the `authorizer`.
```

cip-0112/cip-0112.md
```diff
@@ -629,36 +705,314 @@ data AllocationView = AllocationView with
...
+        --
+        -- They serve as proof that all transfers executed as part of settlement
+        -- were authorized by both sender and receiver.
+        --
+        -- The implementation of this choice determines the validation performed
+        -- by the instrument admin when settling an allocation created by an
+        -- instrument holder.
+
+        -- To minimize the risk of allocations being settled in the wrong
+        -- context, implementations SHOULD ensure that:
```
meiersi-da: should we go for `MUST` here?

cip-0112/cip-0112.md
```diff
@@ -629,36 +705,314 @@ data AllocationView = AllocationView with
...
+
+-- | View of an `AllocationRequest`.
+data AllocationRequestView = AllocationRequestView with
+    ...
+    allocations : [RequestedAllocation]
+      -- ^ The allocations that are requested to be authorized for execution as
+      -- part of the settlement.
+```
+
+If users accept an `AllocationRequest`, Wallets MUST create one `Allocation` per `RequestedAllocation`, and they MUST tidy up the allocation by calling `AllocationRequest_Accept`.
```
meiersi-da:
```suggestion
If users accept an `AllocationRequest`, Wallets MUST create one `Allocation` per `RequestedAllocation`, and they SHOULD tidy up the allocation request by calling `AllocationRequest_Accept`. 
```

I thought we agreed on SHOULD, as it provides more flexibility (e.g., in multi-sync cases), and does not increase the work for the apps, as they anyways have to handle the case of a V1 wallet, which won't archive the request.

cip-0112/cip-0112.md
```diff
@@ -109,18 +111,75 @@ Case in point, CIP-0056 `Allocations` which are core to the settlement flows rel
 
 The token standard V2 proposed here systematically replaces the beneficial owner as represented by a `Party` with an `Account` data structure to capture the information in a standardized way and allow wallets and settlement venues to make use of the information. It also adds flexible authorization to all the interfaces allowing different tokens to encode different authorization rules for holdings held in a traditional account.
 
+### 4.1.3.1 Accounts representing Non-Holder Destinations
+
+Using Accounts rather than `Party` as the source and destination of operations like transfers also allows for the clean representation of non-transfer operations like burns and mints. This is analogous to Ethereum where commonly the `null` address (0x0000000000000000000000000000000000000000) is used as the source/destination for mints and burns. Canton doesn't have a "null Party", but the `Account` data structure can represent special ownerless accounts that represent the source and destination of mints and burns, respectively.
```
meiersi-da:
```suggestion
Using Accounts rather than `Party` as the source and destination of operations like transfers also allows for the clean representation of non-transfer operations like burns and mints. This is analogous to Ethereum where commonly the `null` address (`0x0000000000000000000000000000000000000000`) is used as the source/destination for mints and burns. Canton doesn't have a "null Party", but the `Account` data structure can represent special ownerless accounts that represent the source and destination of mints and burns, respectively.
```

cip-0112/cip-0112.md
```diff
@@ -37,8 +37,10 @@ We propose to evolve the Canton Network token standard defined in [CIP-0056](../
 * `splice-api-token-allocation-allocation-v2`
 * `splice-api-token-holding-v2`
 * `splice-api-token-transfer-instruction-v2`
+* `splice-api-token-standard-utils` for shared types and functions, including those shared between versions and used for cross-version compatibility.
+* `splice-api-token-transfer-events-v2` for the new transaction parsing functionality.
 
-We do not propose to add `splice-api-token-metadata-v2` Daml package, as there is no change required to the v1 version to support the objectives of this CIP. We do however propose a backwards compatible change to the `token-metadata-v1.yaml` file to allow assets to advertise support for non-basic accounts.
+We do not propose to add `splice-api-token-metadata-v2` Daml package, as there is no change required to the v1 version to support the objectives of this CIP. We do however propose a backwards compatible change to the `token-metadata-v1.yaml` OpenAPI specification file to allow assets to advertise support for non-basic accounts and reporting the pause status of an asset.
```
meiersi-da, line 43: AFAIS, the support for `non-basic` accounts is now shown in any section, is it?

cip-0112/cip-0112.md
```diff
@@ -629,36 +705,314 @@ data AllocationView = AllocationView with
...
+```
+
+## 4.3.8 App, Wallet, and Asset Guidelines for Allocations
+
+CIP-0112 sets these guidelines for apps, wallets, and assets on how to map `AllocationRequests` to `Allocations` and settle them:
+
+1. A settlement is specified by an app by setting `SettlementInfo` and a list of `TransferLegs`. The `Accounts` appearing as `sender` or `receiver` on a transfer leg are collectively called the _trading accounts_.
+   * Apps MUST ensure that the `settlementRef` identifier of the settlement is unique for all settlements with the same `executors`, so that the `settlementRef` can be used to reliably match allocations with allocation requests in UIs and backends.
+2. For every trading account, the app SHOULD create one `AllocationRequest` visible to the trading account `owner` and `provider`, which covers *all* transfer legs for the settlement where the trading account appears.
+   * Transfer Legs should be sorted by `admin`, usually into one `RequestedAllocation` per `admin`.
```
meiersi-da:
```suggestion
   * Transfer Legs should be grouped by `admin`, usually into one `RequestedAllocation` per `admin`.
```

cip-0112/cip-0112.md
```diff
@@ -629,36 +705,314 @@ data AllocationView = AllocationView with
...
    * Wallets SHOULD accept the `AllocationRequest` by calling `AllocationRequest_Accept` as part of the allocation transaction to provide replay protection.
-4. `Allocations` MUST NOT contain transfer legs for instruments with different admin parties. Assets MUST ensure that this is not possible.
-5. Assets SHOULD ensure that if there is a set of matching `Allocations`, they are guaranteed to be settleable through the `SettlementFactory_SettleBatch` choice. Matching means identical `SettlementInfo`, and every transfer leg appears once on the sender `Allocation` and once on the receiver `Allocation`.
-   * Exceptions to the settlement guarantee can be made if the context changed between the time of creation of the Allocation and the time of settlement. E.g. it’s acceptable to fail settlement if a trading account was added to a blocklist.
-6. Allocations that lock holdings \- like Canton Coin Allocations \- SHOULD only lock the *net* amounts needed by the trader to fund the settlement.
+4. `Allocations` MUST NOT contain transfer legs for instruments with different `admin` parties. Assets MUST ensure that this is not possible.
```
meiersi-da:
```suggestion
```
No longer required, as the types exclude this by construction.

cip-0112/cip-0112.md
```diff
@@ -629,36 +705,314 @@ data AllocationView = AllocationView with
...
+5. [TODO - This doesn't work] Assets SHOULD ensure that if there is a set of matching `Allocations`, they are guaranteed to be settleable through the `SettlementFactory_SettleBatch` choice. Matching means identical `SettlementInfo`, and every transfer leg appears once on the sender `Allocation` and once on the receiver `Allocation`.
+   * Exceptions to the settlement guarantee can be made if the context changed between the time of creation of the `Allocation` and the time of settlement. E.g. it’s acceptable to fail settlement if a trading account was added to a blocklist.
+6. Allocations that lock holdings, like Canton Coin Allocations, SHOULD only lock the *net* amounts needed by the trader to fund the settlement, including the amounts on `nextIteratedSettlement`.
 
 The reason for these recommendations is that it makes it tractable for wallets’, assets’ and apps’ to perform a complex settlement with privacy and good UX as outlined with 4.1.1 and 4.1.2.
```
meiersi-da, line 1040: Missing: guidance on number of transfer legs.

cip-0112/cip-0112.md
```diff
@@ -771,41 +1125,46 @@ The resulting compatibility matrices are laid out in [5.4](#54-transfer-compatib
...
+    * Only a single `executor` may be set.
+    * `committed` MUST be `False`.
+    * `nextIteratedSettlement` MUST be `None`.
+  * Apps MUST validate that the `Allocations` they got match what they requested and SHOULD reject if they don't. Especially when using new features like iterated settlement, V1 wallets may create `Allocations` that do not match expectations.
 * Assets
   * `TransferInstruction` and `Holding` implementations SHOULD implement both the V1 and V2 interfaces. For transfers that need provider authorization, the existing `TransferInstructionPending` status should be used to communicate extra steps.
     * No matter whether the V1 or V2 interface choice is called, the transaction parsing information suitable for both V1 and V2 transaction parsers MUST be emitted.
       * For V1 transaction parsers, the fallback to metadata is acceptable.
+      * For V2 transfation parsers, every change to holdings (meaning a create or archive of a `Holding`) MUST be represented in exactly one event of type `EventLog_HoldingsChange` with matching `admin` and `Account`.
```
meiersi-da:
```suggestion
      * For V2 transaction parsers, every change to holdings (meaning a create or archive of a `Holding`) MUST be represented in exactly one event of type `EventLog_HoldingsChange` with matching `admin` and `Account`.
```

missing: spell-check over the whole doc. Works pretty well to delegate to Copilot in VS code. Likewise a check that all references are still correct.

cip-0112/cip-0112.md
```diff
@@ -844,53 +1221,47 @@ Legend
...
 Canton Coin is upgraded to support
 
 1. The V2 interfaces in addition to the V1 interfaces.
 2. The backwards compatibility rules of [Section 5](#5-backwards-compatibility)\.
-3. The optimizations 4.3.5.3.
+3. The optimizations 4.3.8.3.
 4. Implicit basic accounts only with `showAccountInputFields` set to `false`.
 
-A first Daml-only draft is available together with the submission of this CIP as part of [the preview branch](https://github.com/hyperledger-labs/splice/blob/meiersi/ts2/preview/token-standard).
+A first Daml-only draft is available together with the submission of this CIP as part of [the preview branch](https://github.com/canton-network/splice/blob/token-standard-v2-upcoming).
```
meiersi-da:
```suggestion
A draft is available together with the submission of this CIP as part of [the preview branch](https://github.com/canton-network/splice/blob/token-standard-v2-upcoming).
```

meiersi-da: we full allocation and transfer v2 support in the amulet wallet UI

meiersi-da (commented):
Nice work! Thanks a lot!

cip-0112/cip-0112.md
bame-da: If you mean "not", not "now", I'm not sure what you mean by "showing the support". The `Account` data structure is shown in 4.3.2. What an asset does with that structure is pretty wide open.

cip-0112/cip-0112.md
bame-da: I decided throughout to put the plural `s` inside the literal strings. For consistency, I'll keep this one as is.

cip-0112/cip-0112.md
bame-da: Yes

cip-0112/cip-0112.md
bame-da: ok

cip-0112/cip-0112.md
meiersi-da: No what I mean is the addition of the `showAccountInputFields` in `token-metadata-v1.yaml` is not explained in this CIP text AFAIS: https://github.com/canton-network/splice/pull/4562/changes#diff-aab06323a36c72e17c72c46e26ab245de2b268b1bf71d9cc6333c99af3e7ee03R170

bame-da: It's in 5.2.

meiersi-da: Oh thanks. Sorry for overlooking that.

cip-0112/cip-0112.md
bame-da: I think SHOULD is right. I'll add the caveat and limits to the guidance section.

cip-0112/cip-0112.md
```diff
@@ -677,39 +677,82 @@ data SettlementInfo = SettlementInfo
...
+```
+data AllocationRequestView = AllocationRequestView with
+    ...
+    requestedAt : Time
+      -- ^ Timestamp at which the request was created.
+    settleAt : Optional Time
+      -- ^ Timestamp at which the settlement is expected to be executed. The authorizer
+      -- SHOULD create their allocations before this time.
+      --
+      -- For iterated settlements, this is the expected time of the first iteration.
```
meiersi-da, line 725: We also need `settlementDeadline` here

cip-0112/cip-0112.md
```diff
@@ -677,39 +677,82 @@ data SettlementInfo = SettlementInfo
...
+      -- authorizer after this time. Thereby recovering storage resources and
+      -- protecting themselves from denial-of-service attacks.
+      --
+      -- Registries SHOULD avoid unnecessary refreshes by
+      --   - making the expiry time as close to the settlement deadline as possible
+      --   - bumping expiry on every settlement iteration.
 ```
 
-Registries SHOULD set `expiresAt` to be equal to `settlementDeadline` if possible, or to the maximum value if not. Value `None` means there is no expiry and users/apps need to tidy up manually.
+Registries SHOULD set `expiresAt` to be equal to `settlementDeadline` if possible, or to the maximum value if not. Value `None` means there is no expiry and users/apps need to tidy up manually.
```
meiersi-da: BTW: consider configuring https://stackoverflow.com/questions/30884131/remove-trailing-spaces-automatically-or-with-a-shortcut/53663494#53663494 to avoid this diff noise

meiersi-da (commented):
ty. looks good, modulo the rename from `SettlementInfo` to `SettlementReference`.

cip-0112/cip-0112.md
bame-da: Shouldn't that go on `RequestedAllocation`?

meiersi-da: apologies. yes, that's where I actually put it. need more ☕

bame-da:
Thanks for all the feedback. Moving to "Ready for review".

waynecollier-da (commented):
My ability to review this doc is limited to finding typos and obvious errors. I am not able to comment on the content itself.

waynecollier-da (approved):

cip-0112/cip-0112.md
```diff
@@ -109,18 +111,79 @@ Case in point, CIP-0056 `Allocations` which are core to the settlement flows rel
...
+`Allocations` in the token standard V1 are intended to represet a a given amount of assets that is guaranteed to be settleable by the `executor` in the context of a broader trade. Atomic settlement is achieved through a four step process:
+
+1. Execute/Agree DvP - transfer legs are now known
+2. Seller allocates exact delivery assets - now guaranteed settleable
+3. Buyer allocates exact payment assets - now guaraneteed settleable
+4. Executor settles - cannot fail
+
+CIP-0112 now adds features that allow important variarions of this flow. 
+
+Firstly, it allows an `Allocation` to be created without the transfer legs already being know. This allows assets to be locked to an app or protocol upfront. For example, this gives the ability to a user to authorize and fund trading of CC vs Stable on an on-chain exchange by creating corresponsing `Allocations`. The exchange can then set the transfer legs after trade execution, as part of settlement. The flow is rearranged to.
```
tiagoaparicio-cpu, line 143:
```suggestion
Firstly, it allows an `Allocation` to be created without the transfer legs already being know. This allows assets to be locked to an app or protocol upfront. For example, this gives the ability to a user to authorize and fund trading of CC vs Stable on an on-chain exchange by creating corresponding `Allocations`. The exchange can then set the transfer legs after trade execution, as part of settlement the flow is rearranged to.
```

PR #240, opened 2026-07-07
Edit: Add description of automated SV weight reduction

waynecollier-da:


waynecollier-da (commented):
During design of the implementation of automated rewards locking it became necessary to spell out the way automated SV weight reductions, and automated Featured App status pauses, would take effect under CIP-0105 and CIP-0105

@hythloda Please submit this and the related PR against CIP-0116 to the SVs for a vote.

PR #245, opened 2026-08-14
CIP: File Governance for Multi-Party Contracts

nandit123:
This PR introduces a new draft CIP: **File Governance for Multi-Party Contracts**. When a Canton contract is shared between two institutions and both reference the same file, there is currently no way for either side, or an auditor, to confirm they are looking at the same file, that neither party altered it, or that it followed its required lifecycle once the deal ended. The CIP documents that gap, sets out where it has already cost institutions money (Tricolor, First Brands, Qingdao Port), surveys existing tooling and where each approach stops, and proposes a boundary between what Canton would track and what stays inside each institution's own systems. It proposes no code and requests no funding: it is posted to establish whether the community considers this a gap worth closing with a shared standard, and lists six open questions for discussion.


4n0nn43x:
Hi Nandit, I read CIP-245 and built a prototype that answers open question 4 (who holds the observer role scoped to detect the same file under two contracts, and under what governance).

The registry is a Decentralized Party on Canton (BitSafe Decentralization Manager, 2-of-3 members), it only ever sees fingerprints (schemaId + hash), and the collision check runs atomically with the lender's token-standard draw: a colliding draw simply does not fund, each lender gets its own notice, an auditor sees collisions and nothing else. It runs end to end on a 3-participant LocalNet today. I can share the repository privately if you want to look at the Daml model.

Two questions for you: which normalised fields would you hash for a loan-pack schema, and is a market-utility party with threshold governance the right answer to question 4? Happy to take feedback here or on a call. We are building this for HackCanton Season 3 and would gladly reference the CIP.

PR #247, opened 2026-08-27
CIP-TBD Remove Whitelisting from Scan and Sequencer on the Global  Synchronizer 

pasindutennage-da:
Fix https://github.com/DACH-NY/canton-network-internal/issues/6662

### DIFF
--- a/cip-01XY/cip-01XY.md
+++ b/cip-01XY/cip-01XY.md
@@ -1,7 +1,7 @@
-# CIP-0101XY
+# CIP-01XY
 
 <pre>
-Number: CIP-0101XY
+Number: CIP-01XY
 Title: Public Sequencer and Scan on the Global Synchronizer
 Author(s):  
   Pasindu Tennage
@@ -37,8 +37,10 @@ License: CC0-1.0
 
 - Two types of revocation (temporary vs. permanent) utilizing the ValidatorUnpermission contract.
 
+- Repermissioning validators once permemently unpermissioned 
 
-### Security
+
+### Deployment Security
 
 - Safety measures: CloudArmor rate limits (per IP, per endpoint, global), and minimal DB user privileges
 
@@ -47,12 +49,13 @@ License: CC0-1.0
 - Prerequisites: Canton 3.X and Splice 0.x.x releases.
 - Phase 1 (Topology Submission): SVs use a feature flag (submitSynchronizerPermission: true) to automatically submit permissions for all existing validators holding MemberTraffic.
 - Phase 2 (Switchover): SVs set a second flag (requireSynchronizerPermissioned: true) to change the Canton DynamicSynchronizerParameters to RestrictedOpen.
-- Handling Unpermissioned Nodes: existing validators without traffic will lose access and must purchase MemberTraffic via standard channels to trigger automatic permissioning.
+- Handling Unpermissioned Nodes: existing validators without MemberTraffic will lose access and must purchase MemberTraffic via standard channels to trigger automatic permissioning.
 
 
 ### DevNet Exception
 
-- specific DevNet endpoint that allows automated self-onboarding and the aggressive IP-based rate limiting applied to it.
+- specific DevNet endpoint that allows automated self-onboarding
+- use of aggressive IP-based rate limiting applied to it.
 
 ### Rollback
 
@@ -61,29 +64,22 @@ License: CC0-1.0
 
 ## Motivation
 
-### Decentralization
-
-- Single SV liability for a validator's behavior is unacceptable. Manual secret generation by sponser SV and scales poorly.
-
-
-### Onboarding Speed
-
-- IP whitelisting is a manual bottleneck, taking up to 4 days to onboard new validators.
-
 ### Governance
 
 - Onboarding and offboarding validators must be a decentralized network decision, not a unilateral one by a single sponsor.
 
+### Onboarding Speed
+
+- Manual secret generation by sponser SV doesn't scale as network grows.
+- IP whitelisting is a manual bottleneck, and doesn't scale when network grows
  
 ## Rationale
 
-- Why Canton 3.5: It natively supports the RestrictedOpen synchronizer state and ParticipantSynchronizerPermission topologies.
+- Why Canton 3.X: It natively supports the RestrictedOpen synchronizer state and ParticipantSynchronizerPermission topologies.
 
 - Why MemberTraffic: It leverages an existing network mechanic for Sybil resistance
 
-- Why Drop IP Whitelists: Because access control is now handled by the on-chain logic, allowing for true public sequencers.
-
-
+- Why drop IP whitelists: Because access control is now handled by the on-chain logic, allowing for true public sequencers.
 
 ## Backwards Compatibility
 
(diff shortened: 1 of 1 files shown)


cip-01XY/cip-01XY.md
```diff
@@ -0,0 +1,106 @@
+# CIP-01XY
+
+<pre>
+Number: CIP-01XY
+Title: Public Sequencer and Scan on the Global Synchronizer
```
martinflorian-da, line 5:
```suggestion
Title: Public Sequencer on the Global Synchronizer
```

Let's not talk about scan here; we discuss it sufficiently in https://github.com/nicu-da/cips/pull/1

I'd also suggest we include "traffic-based onboarding" in the title, but no strong opinion. Something like "Traffic-based onboarding for making sequencers public on the GS"

cip-01XY/cip-01XY.md
```diff
@@ -0,0 +1,106 @@
...
+Status: Pending
+Created: 2026-08-26 
+Approved: 2026-xx-xx 
+License: CC0-1.0
+</pre>
+
+
+## Abstract
+
+- What does "Permissioned Synchronizer" mean for the Canton Network.
```
martinflorian-da, line 19: We should avoid the "permissioned" term; it sends the wrong message to many people that will read this CIP. If we absolutely need to refer to how the Canton feature is called, let's do it, but in all the rest of the text let's talk about "traffic-based onboarding" and "allowing the blocking of nodes at a protocol level".

cip-01XY/cip-01XY.md
```diff
@@ -0,0 +1,106 @@
...
+
+- The transition from IP-whitelist-based access control to permissioned, traffic-based access.
```
martinflorian-da, line 21:
```suggestion
- The transition from IP-whitelist-based access control to traffic-based access.
```

cip-01XY/cip-01XY.md
```diff
@@ -0,0 +1,106 @@
...
+
+- Sybil Resistance; buying minimum MemberTraffic replaces IP whitelisting as the Sybil resistance mechanism.
```
martinflorian-da, line 23: you might need to explain what Sybil resistance means if you insist on using that term

cip-01XY/cip-01XY.md
```diff
@@ -0,0 +1,106 @@
...
+
+- Onboarding and offboarding now require 2f+1 majority SV consensus, removing the unilateral "sponsor SV" model.
```
martinflorian-da, line 25:
```suggestion
```

Kind of already the case / a bit orthogonal IMO.

cip-01XY/cip-01XY.md
```diff
@@ -0,0 +1,106 @@
...
+
+- The proposal includes a one-time manual network switchover from unpermissioned to permissioned.
```
martinflorian-da, line 27:
```suggestion
- The proposal includes a one-time manual network switchover to support the enforcement of traffic-based onboarding.
```

...or something like that

cip-01XY/cip-01XY.md
```diff
@@ -0,0 +1,106 @@
...
+
+
+## Specification
+
+### Architecture
+
+- Brief comparison of existing unilateral SV sponsor model with the new decentralized 2f+1 SV on-chain voting model.
+
```
martinflorian-da, line 35: Here would IMO be the (only) place where we might need to use `permissioned` - to describe Canton's operating modes and how they differ.

cip-01XY/cip-01XY.md
```diff
@@ -0,0 +1,106 @@
...
+- How MemberTraffic triggers the automated issuance of ParticipantSynchronizerPermission.
+
+- Two types of revocation (temporary vs. permanent) utilizing the ValidatorUnpermission contract.
+
+- Repermissioning validators once permemently unpermissioned 
+
+
+### Deployment Security
+
+- Safety measures: CloudArmor rate limits (per IP, per endpoint, global), and minimal DB user privileges
```
martinflorian-da, line 45: Remove this section and instead make a placeholder somewhere that other "drop the whitelists" concerns will be handled in the other CIP: https://github.com/nicu-da/cips/pull/1/changes (maybe 121)

cip-01XY/cip-01XY.md
```diff
@@ -0,0 +1,106 @@
...
+
+### Network Transition
+
+- Prerequisites: Canton 3.X and Splice 0.x.x releases.
```
martinflorian-da, line 49: Is this not TMI for a CIP? Can we know in advance the version numbers?

cip-01XY/cip-01XY.md
```diff
@@ -0,0 +1,106 @@
...
+- Phase 1 (Topology Submission): SVs use a feature flag (submitSynchronizerPermission: true) to automatically submit permissions for all existing validators holding MemberTraffic.
```
martinflorian-da, line 50: How about we make a larger timeline (somewhere in this CIP) that also includes "switching to new onboarding flow" and "deprecating old onboarding flow" ... and yes, why not also "drop the whitelists.

cip-01XY/cip-01XY.md
```diff
@@ -0,0 +1,106 @@
...
+### Rollback
+
+- manual, off-chain SV coordination required to switch the network back to UnrestrictedOpen in case of an emergency.
+
+
+## Motivation
+
+### Governance
+
+- Onboarding and offboarding validators must be a decentralized network decision, not a unilateral one by a single sponsor.
```
martinflorian-da, line 69: You are making it sound as if this was not currently the case... Not sure you need this section tbh.

The main motivation is just "we need to drop whitelists", is it not? No need to motivate the solution in huge detail IMO.

cip-01XY/cip-01XY.md
```diff
@@ -0,0 +1,106 @@
...
+
+### Onboarding Speed
```
martinflorian-da, line 71: The business concerns (we want an open network) are stronger than this; just skip this section IMO.

cip-01XY/cip-01XY.md
pasindutennage-da: Fixed

cip-01XY/cip-01XY.md
pasindutennage-da: Fixed

cip-01XY/cip-01XY.md
pasindutennage-da: fixed

cip-01XY/cip-01XY.md
pasindutennage-da: No, currently only SV  on/off boarding is 2f+1. Here I am refering to validator onboarding which is currently sponser based. Will explain more clearly.

cip-01XY/cip-01XY.md
pasindutennage-da: I saw this mentioned in previous CIPs, and hence added it. We can put real numbers once the implementation release dates are fixed?

cip-01XY/cip-01XY.md
pasindutennage-da: It is indeed not the case now; currently validator onboarding is sposer sv secret based (while sv onboading has 2f+1)

IMO, one main mortivation of the traffic based onboarding was to remove this dependency, and IMO this is new for GS.

cip-01XY/cip-01XY.md
martinflorian-da: Hm and I was thinking about "they need to be whitelisted by 2/3 of SVs". But I guess strictly speaking this is not a must as they can also configure non-BFT scan and sequencer connections...

